PR #34, opened 2020-07-17
fixing issue of setting language flag in ner script

JoPfeiff:
`language` flag causes error when not loading a language adapter. Fixed by no longer setting `language` and instead extracting the language `adapter_name` directly from the adapter config.

### DIFF
--- a/examples/token-classification/run_ner.py
+++ b/examples/token-classification/run_ner.py
@@ -38,9 +38,7 @@ from transformers import (
     set_seed,
     setup_task_adapter_training,
 )
-
 from utils_ner import NerDataset, Split, get_labels
-
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


examples/token-classification/run_ner.py
```diff
@@ -168,13 +169,17 @@ def main():
 
     # Setup adapters
     task_name = "ner"
-    language = adapter_args.language
     setup_task_adapter_training(model, task_name, adapter_args)
     if adapter_args.train_adapter:
+        language = None
+        for k, v in model.config.adapters.adapters.items():
+            if v[0] == "text_lang":
```
calpt, line 176: I'm not sure if the behaviour of this part is intuitive if the loaded model has multiple language adapters since the user cannot specify which one to use. Imho we could just use the existing `adapter_args.load_lang_adapter` flag (maybe rename it) to determine the language here. That flag checks for an existing language adapter and tries to load a pre-trained adapter if none was found. Would that be sufficient?

calpt:
Closing in favor of #52.

PR #37, opened 2020-07-20
unfreezed layernorms of Bert encoders in houlsby architecture

hosein-m:
According to Houlsby's [paper](https://arxiv.org/abs/1902.00751), during adapter tuning, the layer normalization layers of BERT encoders should be trained on the downstream task.


hosein-m:
I found that the trainable layer normalizations in the Housbly's architecture are new and in addition to the BERT's layer normalizations. So, I closed this pull request.

PR #40, opened 2020-07-22
Added multilabel option for classification head

JeroenVanHautte:


src/transformers/adapter_bert.py
```diff
@@ -753,6 +759,17 @@ def forward_head(self, outputs, head_name=None, attention_mask=None, labels=None
                     loss = loss_fct(logits.view(-1, head["num_labels"]), labels.view(-1))
                 outputs = (loss,) + outputs
 
+        elif head["head_type"] == "multilabel_classification":
+            logits = self.heads[head_name](sequence_output[:, 0])
+
+            outputs = (logits,) + outputs[2:]
+            if labels is not None:
+                loss_fct = BCEWithLogitsLoss()
+                if labels.dtype != "torch.float32":
```
JoPfeiff, line 768: I am not sure if this is correct. 
shouldn't it be `labels.dtype != torch.float32`?

JoPfeiff:
Hi Jeroen, thanks for the PR.
I think this is valuable and we should be able to merge it quite fast!

src/transformers/adapter_bert.py
JeroenVanHautte: I double-checked and this was indeed an error, thank you for pointing that out - I have pushed an update resolving it.

JoPfeiff:
+1

PR #47, opened 2020-08-17
Fix config retrieval for adapter fusion & bug in ModelAdaptersConfig

calpt:
This PR includes two fixes:

- Adapter fusion layers do not require identical adapter configs anymore. However, to correctly add fusion layers, all fused adapters still have to define the same _locations_ for adapters (e.g. all should define either ``mh_adapter=True`` or ``mh_adapter=False``). Please check this @arueckle @JoPfeiff 
- Fixed a bug in the ``ModelAdaptersConfig.get()``

Added a test case covering the points mentioned.


arueckle (approved):
looks good + tested that it does not break the inference quickstart (no fusion)

PR #66, opened 2020-09-28
Hotfix for adapter config saving/loading

calpt:
#42 introduced a severe bug when saving & reloading a full model with adapters (ModelAdaptersConfig is not serialized/ deserialized correctly)


arueckle (approved):
👍

PR #70, opened 2020-10-08
Updating contributing.md

NirantK:
The contributing file was inherited as-is from transformers repository. Made some minor edits to the markdown to make it relevant to this repository. I am assuming that the make and test based flows have not changed from the huggingface repo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,12 +131,6 @@ Follow these steps to start contributing:
    it with `pip uninstall transformers` before reinstalling it in editable
    mode with the `-e` flag.)
 
-   Right now, we need an unreleased version of `isort` to avoid a
-   [bug](https://github.com/timothycrosley/isort/pull/1000):
-
-   ```bash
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -214,7 +208,7 @@ CircleCI does not run them.
 
 ### Tests
 
-You can run 🤗 Transformers tests with `unittest` or `pytest`.
+You can run 🤗 Transformers (the underlying library for adapter) tests with `unittest` or `pytest`.
```
calpt, line 211: This would not only run the HuggingFace tests but also some additional tests specific to adapter-transformers. Maybe this formulation is a bit misleading?

calpt (commented):
Thanks for the changes! Just a minor comment.

CONTRIBUTING.md
NirantK: Thanks for catching this @calpt ! 

I've added a short phrase to make it clear that both transformers and adapter-transformer tests will be run.

calpt (approved):
Thanks!

PR #75, opened 2020-10-19
Label Information

hSterz:
The label information is now saved with the head. For a model with a specific head where the head config is None it is saved as an additional parameter. For a model with multiple heads it is a parameter in the ehad config, so every head can have different labels.


calpt (changes requested):
Looks nice overall! 👍 Just a few comments/ questions below.

calpt (approved):

PR #81, opened 2020-11-12
Current id2label in ModelWithHeads config

hSterz:
+ Set the id2label label2id and labels when the active head changes


arueckle:
Thanks! Could you provide a reference to the bug or feature request?

calpt (commented):
Looks good, that should do the trick for the current two cases. One thing to consider: One might want to change the head independent from adding an adapter or calling `set_active_adapters()`. Currently, that can be done just by setting `active_head` directly (not sure, but I think it's also documented like that) which wouldn't switch the map. Two ideas:
1) add a new `set_active_head()` method
2) (preferred) make `active_head` a [python property](https://www.programiz.com/python-programming/property)

calpt (approved):

PR #88, opened 2020-11-24
Split Prediction Heads and make Custom Heads possible

hSterz:
Made individual modules for the prediction heads
Custom Heads can be used and saved that inheret from the PredictionHead

[https://trello.com/c/e8bOC4zZ](url)


calpt (approved):
👍

PR #90, opened 2020-11-25
Adapt flex head model input to HF head model input

calpt:
@hSterz Small changes to label input for our prediction heads to be in line with HF heads (as discussed). Additionally fixed the input to flex head models with multiple choice head. Also moved tests for adapter example scripts to a separate file.

### DIFF
--- a/src/transformers/modeling_bert.py
+++ b/src/transformers/modeling_bert.py
@@ -933,6 +933,11 @@ class BertModelWithHeads(BertModelHeadsMixin, BertPreTrainedModel):
         attention_mask = attention_mask.view(-1, attention_mask.size(-1)) if attention_mask is not None else None
         token_type_ids = token_type_ids.view(-1, token_type_ids.size(-1)) if token_type_ids is not None else None
         position_ids = position_ids.view(-1, position_ids.size(-1)) if position_ids is not None else None
+        inputs_embeds = (
+            inputs_embeds.view(-1, inputs_embeds.size(-2), inputs_embeds.size(-1))
+            if inputs_embeds is not None
+            else None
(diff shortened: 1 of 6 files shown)


src/transformers/modeling_bert.py
```diff
@@ -922,17 +922,22 @@ def forward(
...
     ):
         input_ids = input_ids.view(-1, input_ids.size(-1)) if input_ids is not None else None
         attention_mask = attention_mask.view(-1, attention_mask.size(-1)) if attention_mask is not None else None
         token_type_ids = token_type_ids.view(-1, token_type_ids.size(-1)) if token_type_ids is not None else None
         position_ids = position_ids.view(-1, position_ids.size(-1)) if position_ids is not None else None
+        inputs_embeds = (
+            inputs_embeds.view(-1, inputs_embeds.size(-2), inputs_embeds.size(-1))
+            if inputs_embeds is not None
+            else None
+        )
```
hSterz, line 940: Why is this reformatting now necessary?

hSterz (commented):
Looks good just one question

src/transformers/modeling_bert.py
calpt: I believe it was already necessary before but the change got lost during syncing with HF changes (I think the parameter was introduced in HF 3.x).

calpt: * also, the code is just copied from `BertForMultipleChoice`

hSterz: That makes sense

hSterz (approved):
Thanks now I understand. It looks  good

PR #101, opened 2020-12-07
Fix module registration of new prediction head modules

### LINKED ISSUE #100: All the Colabs with adapter training fail due to tensors being on different devices
Hi,

I'm trying to train an adapter and I'm getting the following error:
`RuntimeError: Tensor for 'out' is on CPU, Tensor for argument #1 'self' is on CPU, but expected them to be on GPU (while checking arguments for addmm)`
to reproduce this, you can just run any of the colabs provided in the tutorials. For example, running the cells in [this](https://colab.research.google.com/github/Adapter-Hub/website/blob/master/app/static/notebooks/Adapter_Quickstart_Training.ipynb#scrollTo=tXHpbKVwwiuz) one, the training line (`trainer.train()`) throws this error.

calpt:
Follow-up to #88. Closes #100.

Currently, the new prediction head modules (`PredictionHead` class) are not correctly registered as submodules of the model by PyTorch as they're not attributes of the model class (which is a PT module) but of the model config (which is not a PT module). This leads to multiple errors (e.g. not part of `state_dict`, ignored when setting train/ eval, moving to GPU (#100)).

Changed the implementation to put head modules into the `heads` module dict in the model class again. `config.prediction_heads` only holds references to the config dicts. Also removed some now unnecessary logic in `adapter_model_mixin.py`.

Added a check for this in prediction heads test cases: https://github.com/calpt/adapter-transformers/blob/3f6c40ca15926e7610aae1904f9b5792851e245d/tests/test_adapter_common.py#L132-L133.

**Edit:**
Added fixes for related issues:
- prediction head for QA (covered by new test in `test_adapter_common.py`)
- restored backwards compability to heads on the Hub by extending PredictionHead from nn.Sequential (covered in `test_adapter_hub.py`)
- cleaned up unused methods

### DIFF
--- a/src/transformers/adapter_bert.py
+++ b/src/transformers/adapter_bert.py
@@ -9,6 +9,7 @@ from .adapter_heads import (
     ClassificationHead,
     MultiLabelClassificationHead,
     MultipleChoiceHead,
+    PredictionHead,
     QuestionAnsweringHead,
     TaggingHead,
 )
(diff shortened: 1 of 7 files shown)


hSterz (approved):
Looks good

PR #103, opened 2020-12-08
 Drop pooler output from flex head output for consistency between BERT & DistilBERT

calpt:
Follow-up of #101 (merge that first).

BERT & RoBERTa models have a pooler module built-in (used for HF prediction heads, but not flex heads). The output of the pooler is returned by the forward() call by default. As other models (e.g. DistilBERT) don't have this pooler, remove it from the input to the prediction head modules for consistent input handling and plug in back afterwards.


hSterz (approved):
Looks fine

PR #107, opened 2020-12-15
Fixed passing labels to model in adapter training notebook

calpt:
Quick fix for the adapter training notebook by explicitly avoiding the label columns to be removed by the HF Trainer. Diff is pretty ugly, the only relevant changes is the last arg in:

```python
training_args = TrainingArguments(
    learning_rate=1e-4,
    num_train_epochs=6,
    per_device_train_batch_size=32,
    per_device_eval_batch_size=32,
    logging_steps=200,
    output_dir="./training_output",
    overwrite_output_dir=True,
    # The next line is important to ensure the dataset labels are properly passed to the model
    remove_unused_columns=False,
)
```

### DIFF
--- a/notebooks/01_Adapter_Training.ipynb
+++ b/notebooks/01_Adapter_Training.ipynb
@@ -15,7 +15,7 @@
     "accelerator": "GPU",
     "widgets": {
       "application/vnd.jupyter.widget-state+json": {
-        "01d575495f3449f1bb900454d851de7c": {
+        "c3aa99f89b76454c96a862454a359f6b": {
           "model_module": "@jupyter-widgets/controls",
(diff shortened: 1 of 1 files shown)


hSterz (commented):
The training seems to work but when I execute this the mapping from id to label is not working correctly. The classifier outputs:
'label': 'LABEL_1'
but as I understand it the model should instead output 👍

calpt:
Not sure if I understood you correctly. What should the classifier output instead of `LABEL_1`?

calpt:
Thanks, great catch! Fixed it.

hSterz (approved):
Now it looks good

PR #110, opened 2020-12-21
Fixes for (re-)loading models with AdapterFusion

calpt:
Fix loading a full model with AdapterFusion (add a test case for this).
Pull _init_adapter_modules() to base mixin to remove redundancy for fusion/ adapter loading.
Add model-specific _add_adapter() method.
Change error for existing fusion config to warning to allow overriding identical fusion.


hSterz (approved):
Looks good

PR #111, opened 2020-12-21
Fix sequence tagging flex head output with given labels.

calpt:
Make sure changes in 34a0ccacdac96d7ce4402b20da1fa6cd3ce99010 work for models with and without prediction heads.
Add prediction head labels to flex head tests.


hSterz (approved):
Looks good to me

PR #116, opened 2021-02-03
Add AdapterDrop to v2

### LINKED ISSUE #84: Integrate AdapterDrop
# 🚀 Feature request

Add our internal implementation of AdapterDrop to the AdapterHub repository.
https://arxiv.org/abs/2010.11918

This consists of several smaller features, which I'll add later to this feature request.

calpt:
Closes #84.

PR is still missing documentation in the docs folder (will be added after docs have been updated to v2).

@arueckle @hSterz please check the comments below.


examples/adapterdrop/drop_at_inference.py
```diff
@@ -0,0 +1,25 @@
+import torch
```
calpt, line 1: @arueckle Wasn't sure what to do with this file. Should we keep (and maybe extend) it or better move the content to the docs or notebooks?

src/transformers/adapter_heads.py
```diff
@@ -335,15 +336,23 @@ def register_custom_head(self, identifier, head):
...
             self.config.id2label = self.get_labels_dict(head_name)
 
+    # TODO: maybe think of something better. remove properties for single head? (breaks compability)
+    @property
+    def active_heads(self):
+        return self._active_heads
+
+    def set_active_heads(self, head_names: Sequence):
+        self._active_heads = head_names
+
```
calpt, line 355: @hSterz For parallel inference, we need to be able to set multiple active heads. This is the simplest possible solution, but having both `active_head` and `active_heads` might be confusing. What do you think?

examples/adapterdrop/drop_at_inference.py
arueckle: We can keep it for now and delete it as soon as we have a colab notebook.

src/transformers/adapter_heads.py
arueckle: I think we need to document what it means to set multiple active heads. There are two options how they could be utilized.

1. Each active head is applied on each input example, resulting in n_examples*K predictions (k being the number of heads)
2. The first head is applied to the first example, second to second ... etc.

arueckle (approved):

src/transformers/adapter_heads.py
calpt: Added some doc to the property.

I also opted for keeping only the `active_head` property (singular) and allowing to pass either a single name or a list. This naming may be a bit weird if passing a list, but it doesn't introduce any further breaking changes to v1 and doesn't have the confusing singular/ plural split.

PR #120, opened 2021-02-17
Enable parallel adapter inference together with attention_mask

calpt:
Follow-up to #116:

In the current parallel inference implementation, it is not possible to set an attention mask at the same time, making parallel inference difficult to use in practice. This PR adds a new hook (`adjust_attention_mask_for_parallel()`) called after every Transformers layer that ensures that the attention mask is properly resized when the hidden states are.

This is enough to make it work for BERT, RoBERTa & DistilBERT, but the Parallel block might not be usable for other model types (e.g. BART). Therefore, I added a whitelist for this block in `adapter_composition.py`.

### DIFF
--- a/tests/test_adapter_composition.py
+++ b/tests/test_adapter_composition.py
@@ -2,16 +2,11 @@ import unittest
 
 import torch
 
-from transformers import (
-    BertConfig,
-    BertForSequenceClassification,
-    BertModelWithHeads,
-    DistilBertModelWithHeads,
-    RobertaModelWithHeads,
-)
-from transformers.adapter_composition import Fuse, Parallel, Split, Stack, parse_composition
+from transformers import AutoModelWithHeads, BertConfig, BertForSequenceClassification
+from transformers.adapter_composition import SUPPORTED_MODELS, Fuse, Parallel, Split, Stack, parse_composition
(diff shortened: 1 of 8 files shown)


tests/test_adapter_composition.py
```diff
@@ -118,6 +118,7 @@ def test_parallel_inference_with_heads(self):
 
             with self.subTest(model_class=model_class.__name__):
                 inputs = {}
+                inputs["attention_mask"] = torch.ones((2, 128))
```
arueckle, line 121: can we have a different attention mask for the different inputs? this would verify whether the attention masks are repeated correctly 

With the current test case we could not detect a difference between the two types of repetitions because a==b
(a,b) -> (a,a,b,b)
(a,b) -> (a,b,a,b)

arueckle (changes requested):
Overall, this looks much cleaner compared to the previous version. 
I think the testcase needs to be adapted slightly to also test for the correctness of the repetition of the attention mask.

tests/test_adapter_composition.py
calpt: Updated to use a random attention mask for each input.

arueckle (approved):

PR #121, opened 2021-02-18
Add guide on how to add adapters to models

calpt:
Drafted a step-by-step guide on how adapters are added to model classes. This might be useful for us to follow when adding new models and it might also serve as a starting point for external contributors.

@hSterz Please check and comment how well this fits with GPT-2.


hSterz (approved):
I think the instructions are very understandable

PR #122, opened 2021-02-24
Support loading adapters with load_best_model_at_end in Trainer

calpt:
Adding `--load_best_model_at_end` currently fails when added to a training script as it tries to load a full model.


hSterz (approved):
Looks good. Maybe a test for this might be usefull

PR #128, opened 2021-03-10
Move adapter loaders & base adapter mixin into separate modules

calpt:
Also improves some doc styles

(This PR doesn't have any changes in logic, only refactorings & docs.)


hSterz (approved):
Looks good

PR #129, opened 2021-03-10
Flex head implementation refactorings

calpt:
Now, the prediction heads supported by a specific model are listed in the `head_types ` dictionary of the heads mixin. The model implementation don't have to implement `add_prediction_head_from_config()` anymore.


hSterz (approved):
Looks good

PR #132, opened 2021-03-24
Add setter for active_adapters property

calpt:
Add setter for active_adapters property.
Set parallel heads automatically with active_adapters.
Fix Parallel with unavailable adapters.


hSterz (approved):
Looks good

PR #133, opened 2021-03-24
Remove InvertibleAdapterConfig class

calpt:
Remove `InvertibleAdapterConfig` class and integrate its keys directly into the base adapter config.
Has addition in `__setattr__` of `AdapterConfig` for backwards compatibility.


hSterz (commented):
Looks good. But it is not clear to me what the advantage of this is

calpt:
Yeah, should've been a bit more clear about this:

The nested InvertibleAdapterConfig was a bit unpractical to use if you would want to change the config, i.e. you would have to first instantiate a new invertible config object class and then assign it to the actual config:
```python
inv_config = InvertibleAdapterConfig(block_type="nice", ...)
config = PfeifferConfig().replace(invertible_adapter=inv_config)
```
Now, you can treat the inv config keys like any other config key:
```python
config = PfeifferConfig().replace(inv_adapter="nice", ...)
```
or
```python
config = AdapterConfig.load("pfeiffer", inv_adapter="nice")
```
(I also think someone brought this up as an issue a long time back :)

hSterz (approved):
Thanks for explaining. Now it makes sense :)

PR #136, opened 2021-03-29
Added new adapter documentation for v2

calpt:
- added document on v1 -> v2 transition
- added document on new composition blocks
- upgraded many things in docs to newest version

@hSterz I have updated everything except for the "Quickstart" and "Adapter Training" pages here, as those are more linked to the notebook/ training scripts updates.


adapter_docs/adapter_composition.md
```diff
@@ -0,0 +1,145 @@
...
+.. important::
+    Fusing adapters with the ``Fuse`` block only works successfully if an adapter fusion layer combining all of the adapters listed in the ``Fuse`` has been added to the model.
+    This can be done either using ``add_fusion()`` or ``load_adapter_fusion()``.
+```
+
+To learn how training an _AdapterFusion_ layer works, check out [this Colab notebook](https://colab.research.google.com/github/Adapter-Hub/adapter-transformers/blob/master/notebooks/03_Adapter_Fusion.ipynb) from the `adapter-transformers` repo.
+
+In v1.x of `adapter-transformers`, fusing adapters was done using a nested list of adapter names, i.e. the example from above would be defined as `[["d", "e", "f"]]`.
+For backwards compatibility, you can still do this, although it is recommended to use the new syntax.
+
```
hSterz, line 80: I think at some point the add_fusion function should be mentioned and explained when to use it.

hSterz (commented):
The different composition blocks are explained very well. Just one comment on that.

adapter_docs/adapter_composition.md
calpt: Actually thought I already did that. Added it now, thanks :)

hSterz (approved):

PR #138, opened 2021-04-01
Dev/trainer checkpoints

hSterz:
Trainer can now resume from checkpoints and saves adapters. 
To do this the do_save_adapters and do_save_full_model parameters need to be set acordingly.


src/transformers/dependency_versions_table.py
```diff
@@ -44,6 +44,7 @@
     "sphinx-copybutton": "sphinx-copybutton",
     "sphinx-markdown-tables": "sphinx-markdown-tables",
     "sphinx-rtd-theme": "sphinx-rtd-theme==0.4.3",
+    "sphinxext-opengraph": "sphinxext-opengraph==0.4.1",
```
calpt, line 47: What's this new dependency?

calpt (approved):
👍

PR #141, opened 2021-04-06
Add adapter support to seq2seq example scripts (translation & summarization)

calpt:
Implementation tested with https://github.com/Adapter-Hub/Hub/pull/24.

This PR also patches the new `check_min_version()` introduced by HF in all example scripts so that it doesn't use our version but the underlying HF version.


hSterz (approved):
Looks good

PR #172, opened 2021-05-12
Allow dynamic dropping of adapter layers in load_adapter()

calpt:


src/transformers/adapters/loading.py
```diff
@@ -408,6 +409,11 @@ def load(
             )
         elif "type" in config:
             self.adapter_type = config["type"]
+        # post-loading drop of layers
+        if leave_out is not None:
+            if config["config"]["leave_out"] is not None:
+                leave_out = list(set(leave_out + config["config"]["leave_out"]))
```
hSterz, line 415:
```suggestion
                # The conversiont to a set and then back to a list removes all duplicates
                leave_out = list(set(leave_out + config["config"]["leave_out"]))
```

hSterz (approved):

PR #183, opened 2021-06-03
Sync/4.6.1

hSterz:
Updating to v4.6.1


calpt (commented):
HF changed the structure of their examples folder such that each supported framework has its own subfolder. I think we could move all our examples to the pytorch folder and remove the other (unsupported) framework folders.

(Also, one file in the examples/legacy folder sneaked in again :)

PR #208, opened 2021-07-19
Prediction head & training example for Dependency Parsing

calpt:
Contains an updated and modified version of the fine-tuning code in https://github.com/Adapter-Hub/hgiyt.

Changes are:
- a new dependency parsing head for BERT & RoBERTa
- an example training script for dependency parsing on Universal Dependencies


examples/dependency-parsing/run_udp.py
```diff
@@ -0,0 +1,342 @@
...
+                reduction_factor=adapter_args.adapter_reduction_factor,
+                leave_out=leave_out,
+            )
+            # load a pre-trained from Hub if specified
+            if adapter_args.load_adapter:
+                model.load_adapter(
+                    adapter_args.load_adapter,
+                    config=adapter_config,
+                    load_as=task_name,
+                    leave_out=leave_out,
```
hSterz, line 204: Why is the leave out specified in the config and during loading?

hSterz (approved):

examples/dependency-parsing/run_udp.py
calpt: There's a slight semantic difference between the two: Setting it in the config searches for adapters that come with the layer dropped, setting it during loading drops it dynamically. I think the first specification is probably not needed.

PR #216, opened 2021-08-06
Improve AdapterFusion config flexibility

calpt:
Move AF config to ModelAdaptersConfig. It's now possible to have multiple AF setups with different configs within one model.

Also removes `set_adapter_fusion_config()`.

### DIFF
--- a/src/transformers/adapters/configuration.py
+++ b/src/transformers/adapters/configuration.py
@@ -195,8 +195,12 @@ class ModelAdaptersConfig(Collection):
         adapters_list = dict(
             map(lambda t: (t[0], t[1][1] or t[1][0] if isinstance(t[1], tuple) else t[1]), adapters_list.items())
         )
-        self.adapters: Mapping[str] = adapters_list
+        self.adapters: Mapping[str, str] = adapters_list
         self.config_map = kwargs.pop("config_map", {})
+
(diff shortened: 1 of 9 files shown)


src/transformers/adapters/configuration.py
```diff
@@ -259,6 +263,55 @@ def add(self, adapter_name: str, config: Optional[Union[str, dict]] = None):
...
+        Adds a new AdapterFusion.
+
+        Args:
+            fusion_name (Union[str, List[str]]): The name of the AdapterFusion or the adapters to fuse.
+            config (Optional[Union[str, dict]], optional): AdapterFusion config. Defaults to None.
+        """
+        if isinstance(fusion_name, list):
+            fusion_name = ",".join(fusion_name)
+        if fusion_name in self.fusions:
+            raise ValueError(f"An AdapterFusion with the name '{fusion_name}' has already been added.")
```
hSterz, line 299: Would an allow overwrite parameter be useful here?

hSterz (approved):
Looks good

src/transformers/adapters/configuration.py
calpt: Overwrites are handled by the methods in `ModelAdaptersMixin` (which are the ones called by the user). The `add()` method for adapters also doesn't allow overwrites here, so keeping it as is.

PR #218, opened 2021-08-10
Separate AdapterTrainer Class

hSterz:
Implement adapter training in a separate class, for better maintenance


calpt (commented):
Looks good overall. Just one question: do we have a test for `load_best_model_at_end` with adapters?

calpt (approved):

PR #225, opened 2021-08-31
Minor improvements for adapter card creation for HF Hub upload

calpt:


hSterz (approved):

src/transformers/adapters/hub_mixin.py
```diff
@@ -1,3 +1,4 @@
+# docstyle-ignore-file
```
hSterz, line 1: Why is this file ignored?

calpt: I didn't want to have the adapter card template strings reformatted

PR #250, opened 2021-11-17
Add `with_head` parameter to save_adapter_fusion

hSterz:
Save head with fusion

### DIFF
--- a/src/transformers/adapters/model_mixin.py
+++ b/src/transformers/adapters/model_mixin.py
@@ -811,12 +811,12 @@ class ModelWithHeadsAdaptersMixin(ModelAdaptersMixin):
         )
 
     def save_adapter_fusion(
-            self,
-            save_directory: str,
-            adapter_names: list,
-            meta_dict: dict = None,
-            custom_weights_loaders: Optional[List[WeightsLoader]] = None,
-            with_head: Optional[str] = False,
+        self,
+        save_directory: str,
+        adapter_names: list,
+        meta_dict: dict = None,
+        custom_weights_loaders: Optional[List[WeightsLoader]] = None,
+        with_head: Optional[str] = False,
     ):
         """
         Saves an adapter and its configuration file to a directory so that it can be shared or reloaded using
--- a/src/transformers/adapters/model_mixin.py
+++ b/src/transformers/adapters/model_mixin.py
@@ -351,7 +351,7 @@ class ModelAdaptersMixin(PushAdapterToHubMixin, ABC):
         Args:
             save_directory (str): Path to a directory where the adapter should be saved.
             adapter_name (str): Name of the adapter to be saved.
-            with_head (str): The name of the head that should be saved with the adapter fusion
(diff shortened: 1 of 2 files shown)


src/transformers/adapters/model_mixin.py
```diff
@@ -469,6 +477,15 @@ def load_adapter_fusion(
...
                 )
+
+        loader = PredictionHeadLoader(self)
+        try:
+            loader.load(load_dir)
+        except ValueError:
+            logger.log(
+                msg="The loaded adapter fusion does not have a head. Before inference you should train or load a head.",
+                level=logging.INFO,
+            )
```
calpt, line 488: Can't we do this similar to `load_adapter()`, i.e. add a custom weights loader in the ModelWithHeadsMixin and set a `with_head` flag there (see l. 759)?

src/transformers/adapters/model_mixin.py
```diff
@@ -350,6 +351,7 @@ def save_adapter_fusion(
         Args:
             save_directory (str): Path to a directory where the adapter should be saved.
             adapter_name (str): Name of the adapter to be saved.
+            with_head (str): The name of the head that should be saved with the adapter fusion
```
calpt, line 354: Parameter name in doc doesn't match real parameter name?

src/transformers/adapters/model_mixin.py
```diff
@@ -362,6 +364,12 @@ def save_adapter_fusion(
             for weights_loader in custom_weights_loaders:
                 weights_loader.save(save_directory, adapter_names)
 
+        if head_name:
+            if head_name not in self.heads:
+                raise ValueError("No head with name {} found".format(head_name))
+            loader = PredictionHeadLoader(self)
+            loader.save(save_directory, head_name)
+
```
calpt, line 372: I assume this would be difficult to do via custom weights loaders 🤔

src/transformers/adapters/model_mixin.py
hSterz: It should be done like that now.

src/transformers/adapters/model_mixin.py
```diff
@@ -810,6 +810,75 @@ def save_all_adapters(
...
+                head with the same name as the adapter fusion is used
+
+        Raises:
+            ValueError: If the given adapter name is invalid.
+        """
+
+        super().save_adapter_fusion(save_directory, adapter_names, meta_dict, custom_weights_loaders)
+
+        if with_head:
+            head_name = with_head if isinstance(with_head, str) else adapter_names
```
calpt, line 838:
```suggestion
            head_name = with_head if isinstance(with_head, str) else adapter_names
            if isinstance(with_head, str):
                head_name = with_head
            elif isinstance(adapter_names, list):
                head_name = ",".join(adapter_names)
            else:
                head_name = adapter_names
```

calpt (approved):
Thanks for including all the feedback :) 

(Realized that the AF methods need some further cleanups in the docs etc. (not related to this directly). Let's do that in a separate PR.)

PR #252, opened 2021-11-24
Unify built-in & custom head implementation

hSterz:
Change adding custom heads to make it more similar to the other heads and more intuitive


calpt (approved):
👍

Issue #165, opened 2021-05-08
Adapter models no longer work effectively in multi-threaded contexts


cody-moveworks:
## Environment info
     
- `adapter-ransformers` version: 2.0.0
- Platform: Darwin-19.6.0-x86_64-i386-64bit
- Python version: 3.6.8
- PyTorch version (GPU?): 1.7.1 (False)
- Tensorflow version (GPU?): 2.4.1 (False)
- Using GPU in script?: No.
- Using distributed or parallel set-up in script?: No.

## Details

With the removal of the `adapter_names` argument in v2.0.0, adapter models no longer work effectively in multi-threaded contexts.

In our environment, we host adapter models in a GRPC server and make requests to the server for predictions from individual adapters. 

In v1.1.0, concurrent calls for predictions from different adapters work as expected since the `adapter_names` are propagated through each layer in the model.

In v2.0.0, we need to set the global state for the base model using `set_active_adapters` method or the `active_adapters` property. This forces us to acquire a lock for the duration of the forward pass and thus we end up having to process the requests for different adapters serially, which is undesirable.

Is it possible to add the `adapter_names` argument back to the forward method signature to support multi-threaded use cases for adapters?

cody-moveworks:
I realize this might actually be more suitably labeled as a bug or a feature request, but I'm unable to change the label.

cc @calpt

calpt:
Hi @cody-moveworks,

Thanks for bringing up this issue. The main motivation for removing the `adapter_names` argument in v2.0 was maintainability as the new approach using `active_adapters` requires much less patches to the HuggingFace code and simplifies synchronizing new upstream changes by them.

However, we were unaware of your usecase with multi-threaded contexts and the unintended consequences of this change. As this definitely seems to be a usecase worth supporting, I think we should re-enable it. I tried to think about solutions that wouldn't require modifying the method signatures of all involved modules, but couldn't find a viable solution (would appreciate it in case you can think of an alternative here). Otherwise, I'll look into restoring the v1.x behavior.

cody-moveworks:
Hi @calpt,

Unfortunately, I also don't see a better solution than modifying the signatures of the involved modules.

Let me know if you need any additional information from me.

Thank you for your consideration on this matter! We greatly appreciate it.

calpt:
We restored the `adapter_names` argument in [v2.0.1](https://github.com/Adapter-Hub/adapter-transformers/releases/tag/adapters2.0.1).

PR #257, opened 2021-12-02
Add `AdapterSetup` context manager

calpt:
## Background

All adapter-supporting model classes currently provide an `adapter_names` parameter in their `forward()` call that allows dynamic & state-less configuration of the activated adapters. We tried to remove this parameter with the release of v2.0.0, however it turned out to be essential for a multi-threaded inference use-case (see #165 for context). The removal was therefore reverted shortly after (#176; v.2.0.1).

## Problem

The `adapter_names` parameter has to be passed down all module classes to the actual adapter layers. This means that the method headers of all top-level and intermediate module classes need to be modified which makes maintanance difficult & prevents further disentangling AH & HF.

## Suggested solution

As a replacement to the `adapter_names` parameter, this PR introduces a new `AdapterSetup` class. This class is intended to be used as a context manager, i.e. a typical use case would look like this:
```python
# will use no adapters
outputs = model(**inputs)

with AdapterSetup(Stack("a", "b")):
    # will use the adapter stack "a" and "b"
    outputs = model(**inputs)
```
Note that in the above example __no__ adapters are activated via `active_adapters`. Within the `with` block, the adapter implementation will read the currently active setup from the context manager.

This solution allows dynamic adapter activation, e.g. also with nesting:
```python
with AdapterSetup(Stack("a", "b")):
    # will use the adapter stack "a" and "b"
    outputs = model(**inputs)
    with AdapterSetup(Fuse("c", "d"), head_setup="e"):
        # will use fusion between "c" and "d" & head "e"
        outputs = model(**inputs)
```

Most importantly, the context manager is **thread-local**, i.e. we can use different setups in different threads with the same model instance. There's a test for this:
https://github.com/calpt/adapter-transformers/blob/93065b74ae933b34ec5c3f4047223792a795239f/tests/test_adapter_setup_context.py#L75-L112

## Breaking changes
- The `adapter_names` parameter is removed for all model classes


src/transformers/models/bart/modeling_bart.py
```diff
@@ -1299,41 +1291,41 @@ def forward(
...
-
-        head_outputs = self.forward_head(
-            outputs,
-            head_name=head,
-            cls_output=cls_representation,
-            attention_mask=attention_mask,
-            return_dict=return_dict,
-            **kwargs,
-        )
+        with AdapterSetup(adapter_names, ignore_empty=True):
```
hSterz, line 1294: Maybe we should remove the `adapter_names` parameter to avoid confusion.

adapter_docs/adapter_composition.md
```diff
@@ -11,6 +11,7 @@ model.active_adapters = "adapter_name"
 ```
 
 Note that we also could have used `model.set_active_adapters("adapter_name")` which does the same.
+Alternatively, the [`AdapterSetup`](classes/adapter_config.html#transformers.AdapterSetup) context manager allows dynamic configuration of activated setups without changing the model state.
```
hSterz, line 14: Maybe an example code snippet would be useful here.

src/transformers/adapters/heads/base.py
```diff
@@ -653,11 +637,17 @@ def delete_head(self, head_name: str):
...
-        if not head_name and not self.active_head:
+        if head_name:
+            used_heads = [head_name]
+        elif AdapterSetup.get_context_head_setup():
+            used_heads = AdapterSetup.get_context_head_setup()
+            if isinstance(used_heads, str):
+                used_heads = [used_heads]
+        elif self._active_heads:
+            used_heads = self._active_heads
+        else:
```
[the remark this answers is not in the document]
calpt: That's right, done.

adapter_docs/adapter_composition.md
calpt: done

src/transformers/models/bart/modeling_bart.py
calpt: done

hSterz (approved):

PR #259, opened 2021-12-08
Fix documentation and consistency issues for AdapterFusion methods

calpt:
Follow-up to #250.


src/transformers/adapters/model_mixin.py
```diff
@@ -339,28 +345,34 @@ def save_adapter(
...
 
         Raises:
-            ValueError: If the given adapter name is invalid.
+            ValueError: If the given AdapterFusion name is invalid.
         """
+        if isinstance(adapter_names, Fuse):
+            adapter_fusion_name = ",".join(adapter_names.children)
+        elif isinstance(adapter_names, list):
+            adapter_fusion_name = ",".join(adapter_names)
+        else:
```
hSterz, line 367: Maybe it would make sense to check whether this is really a string in the else case and throw an error otherwise

hSterz (approved):
Looks good.

PR #228, opened 2021-09-03
Ensure reference from adapter layer to layer norm is always correct

calpt:
Fixes errors with torch DataParallel.
Closes #227.

PR #263, opened 2021-12-16
Unify & simplify model adapter integration

calpt:
**Note: Depends on #257**

---

## Background

Currently, we have a lot of redundant code in the model mixins of each model class integration (`src/transformers/adapters/models`).

## Solution

This PR does a range of refactoring with the goal to unify and simplify adapter the adapter integrations into the currently supported models:

- `AdapterLayerBaseMixin` has been converted to a torch module, `AdapterLayer`, with a proper `forward()` method. This makes the integration into most models cleaner (except for (Ro)BERT(a)/ T5) as they don't require specific module mixins anymore
- The transformer now is expected to be passed to `AdapterLayer` instead of being a class attribute. This removes the need for a couple of model integration workarounds.
- A lot of management code (`train_adapter()`, `train_adapter_fusion()`, `add_adapter()`, `get_adapter()` etc.) has been pulled up to `ModelAdaptersMixin`. All of these methods now are model-agnostic and have been removed from each model-specific mixin.
- To enable the last point, a new abstract `iter_layers()` method has been introduced in `ModelAdaptersMixin`. In the best case (e.g. DistilBERT), this is the **only** method that **has to be implemented** by each model mixin.
- A new mechanism to call methods on each `AdapterLayer` has been added (`ModelAdaptersMixin.apply_to_adapter_layers()`). This means a base model mixin can _directly_ call `AdapterLayer` methods. All intermediate `add_adapter()`, `enable_adapters()` etc. methods in each model layer are not necessary anymore.
- As a consequence of the last points, all module mixins between the module holding `AdapterLayer` and the base model mixin have been removed (e.g. `BertEncoderAdaptersMixin`).
- All tensor adjustments for parallel adapter composition now use `adjust_tensors_for_parallel()` in `composition.py`.

## Breaking changes

- The return of `get_adapter()` has changed.

### DIFF
--- a/src/transformers/adapters/layer.py
+++ b/src/transformers/adapters/layer.py
@@ -1,4 +1,3 @@
-from abc import ABC, abstractmethod
 from typing import List, Mapping, Union
 
 import torch
@@ -9,22 +8,11 @@ from .context import AdapterSetup
 from .modeling import Adapter, BertFusion
 
 
-class AdapterLayerBaseMixin(ABC):
-    """
-    An abstract base implementation of adapter integration into a Transformer block. In BERT, subclasses of this module
-    are placed in the BertSelfOutput module and in the BertOutput module.
-    """
-
-    # override this property if layer norm has a different name
-    @property
-    def transformer_layer_norm(self):
-        return self.LayerNorm
-
-    @property
-    @abstractmethod
-    def adapter_config_key(self):
-        """Gets the name of the key by which this adapter location is identified in the adapter configuration."""
-        pass
+class AdapterLayer(nn.Module):
+    def __init__(self, location_key: str, config):
+        super().__init__()
+        self.location_key = location_key
+        self.config = config
 
     @property
     def layer_idx(self):
@@ -43,7 +31,7 @@ class AdapterLayerBaseMixin(ABC):
     def add_adapter(self, adapter_name: str, layer_idx: int):
         self.layer_idx = layer_idx
         adapter_config = self.config.adapters.get(adapter_name)
-        if adapter_config and adapter_config.get(self.adapter_config_key, None):
+        if adapter_config and adapter_config.get(self.location_key, None):
             reduction_factor = adapter_config["reduction_factor"]
             if isinstance(reduction_factor, Mapping):
                 if str(self.layer_idx) in reduction_factor:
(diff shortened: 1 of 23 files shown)


adapter-hub-bert:
This PR/issue depends on:
* ~~Adapter-Hub/adapter-transformers#257~~
---
Automatically provided by **[Dependent Issues](https://github.com/z0al/dependent-issues)** (🤖).

src/transformers/adapters/layer.py
```diff
@@ -449,30 +470,37 @@ def adapters_forward(self, hidden_states, input_tensor):
...
-        elif self.transformer_layer_norm:
-            hidden_states = self.transformer_layer_norm(hidden_states + input_tensor)
+        elif layer_norm:
+            hidden_states = layer_norm(hidden_states + input_tensor)
         else:
             hidden_states = hidden_states + input_tensor
 
         return hidden_states
+
+    def forward(self, hidden_states, input_tensor, layer_norm):
```
hSterz, line 505: What is the advantage of passing that with every forward call instead of having a parameter in the class containing the layer norm?

src/transformers/adapters/models/bert.py
```diff
@@ -15,162 +14,35 @@
...
+from ..layer import AdapterLayer
 from ..model_mixin import InvertibleAdaptersMixin, ModelAdaptersMixin
 
 
 logger = logging.getLogger(__name__)
 
 
-class BertSelfOutputAdaptersMixin(AdapterLayerBaseMixin):
+# For backwards compatibility, BertSelfOutput inherits directly from AdapterLayer
+class BertSelfOutputAdaptersMixin(AdapterLayer):
```
hSterz, line 25: Why do some models still have AdapterMixins and others don't e.g. in the BART implementation, the `BartSelfAttentionAdaptersModule` is completely removed?

hSterz (commented):
This looks like some great changes that will make things easier. I just have some small questions :)

src/transformers/adapters/layer.py
calpt: The layer norm usually is a child module of the Transformer block module (or part of it as e.g. for BERT). As the `AdapterLayer` usually also is a child module of the same block module, we would have to keep a reference to an attribute of the parent module. As this makes the implementation less clean and poses some issues (cf. #228), I decided to pass the layer norm in the forward pass.

I might have missed some better solution though, please let me know if there are any :)

src/transformers/adapters/models/bert.py
calpt: Mainly for compatibility reasons. I didn't want to change the weight names and write yet another weight renaming method, so mixins are kept wherever necessary.

calpt: However, we could go even further and also remove mixins such as `BartEncoderLayerAdaptersMixin` and just add the code to the main modeling classes as they mainly consist of creating child modules 🤔

PR #265, opened 2021-12-27
fix small typo

ramitsurana:
Hi,

I was experimenting with adapters and found a minor typo. Creating the PR for the same.

Thanks


calpt (approved):
Great, thanks for fixing it!

PR #298, opened 2022-03-09
Rename non-linearities to fit huggingface conventions

### LINKED ISSUE #291: activation functions
It would be nice if the activation function names were identical to those in huggingface's activations, and also include all the activations that reside in there, such as quick_gelu, etc.  Not clear as to why a different mapping is required?

hSterz:
Change the name of the activation functions to match the ones by huggingface (Resolves #291 ):

-  change `gelu` to `gelu_new` and `gelu_orig` to `gelu`
- handle loading adapters saved before the name change
- save the `__adapter_version__` with the adapter


src/transformers/adapters/loading.py
```diff
@@ -52,6 +55,8 @@ def save_weights_config(self, save_directory, config, meta_dict=None):
             for k, v in meta_dict.items():
                 if k not in config:
                     config[k] = v
+        # add version to config
+        config["adapter_version"] = __adapters_version__
```
calpt, line 59: Would it make sense to move this to the `build_full_config()` method in configuration.py as similar data is added over there?

PR #305, opened 2022-03-15
Add XLM-Roberta to list of models supporting `Parallel`

### LINKED ISSUE #282: Why is Pararallel Composition blocked for XLM-Roberta
Parallel composition of adapters are only supported for a few model types listed here:
https://github.com/Adapter-Hub/adapter-transformers/blob/9a6bf1757b684a4c627c5a35a56e61ea706dccee/src/transformers/adapters/composition.py#L101-L103

But I believe XLM-Roberta has the same architecture as Roberta, so I think parallel composition will also work with XLM-Roberta models. Is it possible to use parallel composition with XLM-Roberta models?

### LINKED ISSUE #294: XLM-R with the Parallel Adapter Composition
Hi, I'm getting the following error when using **XLM-R** with the **parallel** composition.

> ValueError: Models of type xlm-roberta don't support adapter composition using Parallel.

But when I am using a parallel block inside a stack, I don't get that error.

Can someone explain the reason? Does parallel composition really not support XLM-R?

calpt:
Closes #282, closes #294

### DIFF
--- a/src/transformers/adapters/composition.py
+++ b/src/transformers/adapters/composition.py
@@ -99,7 +99,7 @@ ALLOWED_NESTINGS = {
 # Some composition blocks might not be supported by all models.
 # Add a whitelist of models for those here.
 SUPPORTED_MODELS = {
-    Parallel: ["bert", "roberta", "distilbert", "bart", "mbart", "gpt2", "t5"],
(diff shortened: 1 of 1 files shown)


gregor-ge (approved):
Tested and it does what it should.

PR #309, opened 2022-03-21
Allow `source=None` to search all source in `load_adapter()`

calpt:


src/transformers/adapters/utils.py
```diff
@@ -466,6 +471,24 @@ def resolve_adapter_path(
...
+            logger.info("Attempting to load adapter from source 'ah'...")
+            return pull_from_hub(
+                adapter_name_or_path, model_name, adapter_config=adapter_config, version=version, **kwargs
+            )
+        except EnvironmentError as ex:
+            logger.info(ex)
+            logger.info("Attempting to load adapter from source 'hf'...")
+            try:
+                return pull_from_hf_model_hub(adapter_name_or_path, version=version, **kwargs)
+            except Exception as ex:
```
hSterz, line 485: Would it make sense to also scan the local file system? Or with which source can a locally stored adapter be loaded?

calpt: Local file system is always checked at the beginning, no matter of source (cf. l. 459)

hSterz: Makes sense

hSterz (approved):

PR #317, opened 2022-03-31
adapters-trainer: add missing preprocess_logits_for_metrics argument

stefan-it:
Hi,

I just wanted to train my first language adapter with the `run_mlm.py` example script. I also used a validation corpus, but training does not start due to this error message:

```bash
Traceback (most recent call last):                                                                                                                                            
  File "run_mlm.py", line 632, in <module>                                                                                                                                    
    main()                                                                                                                                                                    
  File "run_mlm.py", line 563, in main                                                                                                                                        
    trainer = trainer_class(                                                                        
TypeError: __init__() got an unexpected keyword argument 'preprocess_logits_for_metrics'
```

I did some debugging and the `preprocess_logits_for_metrics` argument is available for the upstream Transformers Trainer interface:

https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L291

But is currently not available for the Adapters Trainer interface. This PR adds the `preprocess_logits_for_metrics` argument to the interface. Training is then working :hugs:


calpt (approved):
Thanks!

PR #325, opened 2022-04-12
Fix save_all_adapters such that with_head is not ignored

### LINKED ISSUE #320: Bug of ModelWithHeadsAdaptersMixin.save_all_adapters, forcedly save heads
In src/transformers/adapters/model_mixin.py, line 560:
```
    def save_all_adapters(
        self,
        save_directory: str,
        meta_dict: dict = None,
        custom_weights_loaders: Optional[List[WeightsLoader]] = None,
    ):
        """
        Saves all adapters of this model together with their configuration to subfolders of the given location.

        Args:
            save_directory (str): Path to a directory where the adapters should be saved.
        """
        for name in self.config.adapters:
            adapter_config = self.config.adapters.get(name)
            h = get_adapter_config_hash(adapter_config)
            save_path = join(save_directory, name)
            if meta_dict:
                meta_dict.update({"config_id": h})
            else:
                meta_dict = {"config_id": h}
            self.save_adapter(save_path, name, meta_dict=meta_dict, custom_weights_loaders=custom_weights_loaders)#The bug
```
when you calling:
```
ModelWithHeadsAdaptersMixin.save_all_adapters(...,with_head=False)
```
It will actually save heads, because in **ModelAdaptersMixin.save_all_adapters** you call **self.save_adapter**, which will not enter **ModelAdaptersMixin.save_adapter** directly but enter **ModelWithHeadsAdaptersMixin.save_adapter** first, and the argument **with_head** is set to **True** by default.

hSterz:
Closes #320

### DIFF
--- a/src/transformers/adapters/model_mixin.py
+++ b/src/transformers/adapters/model_mixin.py
@@ -931,15 +931,15 @@ class ModelWithHeadsAdaptersMixin(ModelAdaptersMixin):
         meta_dict: dict = None,
         custom_weights_loaders: Optional[List[WeightsLoader]] = None,
     ):
-        if with_head:
-            if custom_weights_loaders is None:
-                custom_weights_loaders = []
-            custom_weights_loaders.append(PredictionHeadLoader(self, error_on_missing=False))
-        super().save_all_adapters(
-            save_directory,
-            meta_dict=meta_dict,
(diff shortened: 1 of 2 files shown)


calpt (approved):
👍

PR #330, opened 2022-04-20
notebook fix

mohdsanadzakirizvi:
The tokenizer code was throwing the error: `AttributeError: 'DatasetDict' object has no attribute 'rename_column_'` which is probably an API change. Made changes in the cell to make it work and the entire notebook ran perfectly.


calpt (approved):
Thanks!

Issue #358, opened 2022-06-03
prefix-tuning does not with pytorch >= 1.5.0


alexanderhanboli:
## Environment info

     
- `adapter-transformers` version: 3.0.1
- Platform: Linux-4.14.256-197.484.amzn2.x86_64-x86_64-with-glibc2.10
- Python version: 3.8.5
- PyTorch version (GPU?): 1.11.0+cu102 (True)
- Tensorflow version (GPU?): not installed (NA)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Using GPU in script?: yes
- Using distributed or parallel set-up in script?: parallel

## Information

Model I am using (Bert, XLNet ...): T5

Language I am using the model on (English, Chinese ...): English

Adapter setup I am using (if any): prefix-tuning

The problem arises when using:
* [X] the official example scripts: (give details below)
* [ ] my own modified scripts: (give details below)

The tasks I am working on is:
* [ ] an official GLUE/SQUaD task: (give the name)
* [X] my own task or dataset: (give details below)

## To reproduce

1. export CUDA_VISIBLE_DEVICES=0,1,2,3 python examples/pytorch/summarization/run_summarization.py --model_name_or_path "google/t5-v1_1-base" --do_train --train_adapter --adapter_config prefix_config.json

This is the prefix_config.json
```
{
    "architecture": "prefix_tuning",
    "encoder_prefix": true,
    "cross_prefix": true,
    "leave_out": [],
    "flat": false,
    "prefix_length": 30,
    "bottleneck_size": 512,
    "non_linearity": "tanh",
    "dropout": 0.0
}
```

This line (https://github.com/adapter-hub/adapter-transformers/blob/master/src/transformers/adapters/prefix_tuning.py#L49) does not work for data parallel. We need to update it to be something similar to (https://github.com/huggingface/transformers/blob/1c220ced8ecc5f12bc979239aa648747411f9fc4/src/transformers/modeling_utils.py#L121).

PR #359, opened 2022-06-03
multi-gpu support for prefix-tuning

alexanderhanboli:
The current way of initializing `input_tokens` in `__init__` and using `device = next(self.parameters()).device` to find device does not work for multi-gpu.

Instead we should use this [fix](https://github.com/huggingface/transformers/blob/1c220ced8ecc5f12bc979239aa648747411f9fc4/src/transformers/modeling_utils.py#L120) to find the device of the module and assign it to the input tokens.

### DIFF
--- a/src/transformers/adapters/prefix_tuning.py
+++ b/src/transformers/adapters/prefix_tuning.py
@@ -3,12 +3,12 @@ from typing import List, Union
 import torch
 from torch import nn
 
+from ..modeling_utils import ModuleUtilsMixin
 from .composition import AdapterCompositionBlock
 from .configuration import PrefixTuningConfig
 from .context import AdapterSetup, ForwardContext
 from .layer import AdapterLayerBase
 from .modeling import Activation_Function_Class
-from ..modeling_utils import ModuleUtilsMixin
 
 
 class PrefixTuning(nn.Module, ModuleUtilsMixin):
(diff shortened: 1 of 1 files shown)


src/transformers/adapters/prefix_tuning.py
```diff
@@ -80,12 +80,11 @@ def __init__(
         self.dropout = nn.Dropout(self.config.dropout)
 
     def forward(self, batch_size):
-        device = next(self.parameters()).device
         key_values = (
             self.control_trans.unsqueeze(0)
             .expand(batch_size, -1)
             .view(batch_size, self.config.prefix_length, self.n_layers * 2, self.n_heads, self.n_embd_per_head)
-            .to(device)
+            .to(self.device)
```
calpt, line 87: Please also add the mixin to the `FlatPrefixTuning` class to pass the tests.

calpt (commented):
Hey @alexanderhanboli, thanks for looking into this issue and providing a fix! Happy to merge this as soon as all the tests are passing.

(I'm assuming this closes #358)

src/transformers/adapters/prefix_tuning.py
alexanderhanboli: Hey @calpt sorry for the delay. I just added the mixin to that class.

calpt (approved):
Looks good, thanks!

PR #426, opened 2022-09-26
Adding adapters support for GPT-J

ChiragBSavani:
Support of adapters for GPT-J with tests and documentation.


calpt (commented):
Hey @ChiragBSavani, thank you very much for your contribution! This looks very good to me, just one tiny final comment: Please also add the GPT-J doc to the table of contents in the adapter_docs `index.rst`.

ChiragBSavani:
@calpt Thank you for looking into the PR. I have made changes according to your comment and also parallel I have fixed one issue which I found out meanwhile. The rest looks very good. 

Please have a look at it again when you have time.

src/transformers/models/gptj/modeling_gptj.py
```diff
@@ -319,7 +332,9 @@ def forward(
         outputs = attn_outputs[1:]
 
         feed_forward_hidden_states = self.mlp(hidden_states)
-        hidden_states = attn_output + feed_forward_hidden_states + residual
+        hidden_states = self.attention_adapters(attn_output, residual, None)
```
calpt, line 335: I believe the call to `attention_adapters()` should be placed before the forward pass through `mlp()` here. (Similar to the implementation in `GPT2Block` in `modeling_gpt2.py`.

calpt: I see, compared with other Transformer implementations, GPT-J doesn't place a separate residual around the attention block. What I was concerned about here was that now, in contrast to the typical adapter placement, the MLP layer does not use the outputs of the attention adapter.

ChiragBSavani: Correct. That is why it is different in GPT-J

calpt: We'll keep it as you suggested then

calpt (approved):

PR #433, opened 2022-10-09
GH Action stale: never stale enhancements

lenglaender:
- doesn't stale enhancement issues
- more operations per run (= more github API calls per run)
- rescheduled action to run at 6:00 UTC


.github/workflows/stale.yml
```diff
@@ -14,10 +14,11 @@ jobs:
     steps:
       - uses: actions/stale@v6
         with:
```
calpt, line 16:
```suggestion
        with:
          repo-token: "${{ secrets.BOT_TOKEN }}"
```

This change lets the stale issue comments be posted under the account of our adapter-hub-bert bot. Not really required, but we do this throughout all repos (cf. e.g. https://github.com/adapter-hub/Hub/blob/master/.github/workflows/label.yml)

calpt (commented):
One additional suggestion, rest looks good.

calpt (approved):

PR #434, opened 2022-10-13
Load sample CIFAR10 data for tests from repo

calpt:


tests_adapters/test_adapter.py
```diff
@@ -96,9 +96,7 @@ def transform(example_batch):
             return inputs
 
         dataset_builder = datasets.load_dataset_builder("cifar10")
-
-        mock_dl_manager = MockDownloadManager("cifar10", dataset_builder.config, datasets.Version("1.0.0"))
-        dataset_builder.download_and_prepare(dl_manager=mock_dl_manager, ignore_verifications=True)
+        dataset_builder.base_path = "tests_adapters/fixtures/samples/cifar10"
```
lenglaender, line 100: I ran the tests with the changes you made and 21 tests still failed. I think we still need to call the download_and_prepare() method. With this, only 3 tests failed:

- tests_adapters/test_vit.py::ViTAdapterTest::test_batch_split_training
- tests_adapters/test_vit.py::ViTAdapterTest::test_parallel_training
- tests_adapters/test_vit.py::ViTAdapterTest::test_train_adapter_fusion

Can you look into this further @calpt?

```suggestion
        dataset_builder.download_and_prepare()
        
```

calpt: Thanks for looking into this. My original changes didn't actually do what was intended, that was a mistake from my side. I've now pushed a new solution including a full HF datasets loader script that passes all tests.

lenglaender (approved):

PR #436, opened 2022-10-21
Enable parallel sequence generation with adapters

calpt:
Sequence generation using the `model.generate()` method currently doesn't support `Parallel` composition of adapters. This PR enables this, i.e. parallel sequence generation using different adapters using `Parallel` block for `AdapterModel` classes and adapters w. seq2seq heads.

Test: https://github.com/calpt/adapter-transformers/blob/827c627ddce5fcec1bef82e33721bcd12933fd51/tests_adapters/test_adapter_composition.py#L249-L274

### DIFF
--- a/tests_adapters/test_adapter_composition.py
+++ b/tests_adapters/test_adapter_composition.py
@@ -15,6 +15,7 @@ from transformers import (
     Trainer,
     TrainingArguments,
 )
+from transformers.adapters import ADAPTER_MODEL_MAPPING
 from transformers.adapters.composition import BatchSplit, Fuse, Parallel, Split, Stack, parse_composition
 from transformers.testing_utils import require_torch, torch_device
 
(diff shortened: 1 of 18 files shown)


tests_adapters/test_adapter_composition.py
```diff
@@ -245,6 +246,33 @@ def test_batch_split_with_heads(self):
...
+
+        model1 = AutoAdapterModel.from_config(self.config())
+        model1.add_adapter("adapter1")
+        model1.add_adapter("adapter2")
+        if hasattr(model1, "add_seq2seq_lm_head"):
+            model1.add_seq2seq_lm_head("adapter1")
+            model1.add_seq2seq_lm_head("adapter2")
+        else:
+            model1.add_causal_lm_head("adapter1")
+            model1.add_causal_lm_head("adapter2")
```
lenglaender, line 264: If a model has both a seq2seq_lm_head and a causal_lm_head, the test only checks if `model.generate()` works with the seq2seq_lm_head, but not if it works with the causal_lm_head.
However, I'm not sure if it's necessary to test both heads.

lenglaender (approved):
Everything else looks good to me

tests_adapters/test_adapter_composition.py
calpt: I think we currently don't have a model that can have both heads. Either the model is encoder-decoder, then it has a seq2seq generation head, or it is decoder-only, then it has a causal generation head.

lenglaender: Ok, then it is most likely not necessary to test both cases.

PR #451, opened 2022-11-09
Fix DeBERTa prefix tuning w. enabled relative attention

calpt:
Current implementation of prefix tuning is incompatible with relative attention of DeBERTa/ DeBERTa v2 (`relative_attention=True`), i.e. throws errors on position-to-content & content-to-position attentions. This PR removes prefix tokens from relative attention computation.


hSterz (approved):
Looks good 👍

PR #473, opened 2023-01-11
Make HuggingFace Hub Mixin work with newer utilities

Helw150:
In the current master branch, this mixin fails since it's using utilities that no longer exist in Transformers itself. I've updated the Adapter HuggingFace Hub mixin to be compliant by following the update PR from the main transformers branch: https://github.com/huggingface/transformers/pull/18366/files


src/transformers/adapters/hub_mixin.py
```diff
@@ -162,54 +162,40 @@ def push_adapter_to_hub(
...
 
         Returns:
             str: The url of the adapter repository on the model hub.
         """
-        repo_url = self._get_repo_url_from_name(
-            repo_name, organization=organization, private=private, use_auth_token=use_auth_token
-        )
-        if local_path is not None:
-            repo_path = local_path
+        if not repo_name.startswith(organization):
```
calpt, line 171: This line throws an exception for me when `organization` is not set.
```suggestion
        if organization is not None and not repo_name.startswith(organization):
```

Maybe we could also think about deprecating the `organization` parameter as done in Transformers?

calpt (commented):
Thanks a lot for looking into this!

Found one issue when testing the code (commented below).

src/transformers/adapters/hub_mixin.py
Helw150: I think deprecating organization to keep the AdapterHub API similar to HuggingFace makes a lot of sense! I just didn't want to change your API as a first time contributor!

Helw150: For now, I've added your suggested change and kept everything backwards compatible. If you want to make this a breaking change and deprecate the organization argument, I can make that shift instead!

calpt: I think we can go for the change but keep the organization parameter for backward compatibility with a deprecation warning, similar to HF does it here: https://github.com/huggingface/transformers/blob/d1c9242db2092e9d96b1b18ba9fe27f3518184a5/src/transformers/utils/hub.py#L891-L899

lenglaender: Hey @Helw150, are you working on the deprecation warning implementation suggested by @calpt? We are planning to release the next AdapterHub version next week and would like to include your pull request.

Helw150: Sorry - hadn't seen this notification. Just added the warning!

calpt (approved):
Thanks again for this, looks good!

PR #476, opened 2023-01-12
Add support for `Stack`, `Parallel` & `BatchSplit` composition to prefix tuning

calpt:
This PR:
- implements `Stack`, `Parallel` & `BatchSplit` composition for prefix tuning modules
- refactors the adapter composition tests to also test adapter composition using prefix tuning


calpt:
Depends on #479.

hSterz (commented):
Looks good to me. We could maybe update the documentation on composition blocks to include prefix tuning.

hSterz (approved):
Looks good to me

PR #479, opened 2023-01-16
Fix T5 adapter layer input

calpt:
The adapter implementation of T5 currently passes the inputs to the adapter layer in the wrong order (i.e. residual inputs before hidden states). Thus, the sequential adapter implementation is effectively a parallel adapter.

This PR fixes the inputs order. It also clarifies the adapter layer input parameters.

### DIFF
--- a/src/transformers/models/t5/modeling_t5.py
+++ b/src/transformers/models/t5/modeling_t5.py
@@ -340,7 +340,9 @@ class T5LayerFF(T5FFLayerAdaptersMixin, nn.Module):
     def forward(self, hidden_states):
         forwarded_states = self.layer_norm(hidden_states)
         forwarded_states = self.DenseReluDense(forwarded_states)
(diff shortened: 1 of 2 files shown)


src/transformers/models/t5/modeling_t5.py
```diff
@@ -340,7 +340,7 @@ def __init__(self, config: T5Config):
     def forward(self, hidden_states):
         forwarded_states = self.layer_norm(hidden_states)
         forwarded_states = self.DenseReluDense(forwarded_states)
-        hidden_states = self.adapter_layer_forward(hidden_states, self.dropout(forwarded_states), None)
+        hidden_states = self.adapter_layer_forward(self.dropout(forwarded_states), hidden_states, None)
```
hSterz, line 343: We could make it super obvious which one the residual connection is by using `hidden_states=..., residual_connection=...` in the method call. This could help people adding new models when they have a look at the currently implemented models.

hSterz (approved):

Issue #481, opened 2023-01-19
AdapterTrainer requires <model> even if <model_init> was provided  


bartoszpawlowicz:
## Environment info

- `adapter-transformers` version: 3.1.0

## To reproduce

1. Create a <model_init> function to be used in hyperparameter tuning
2. Pass a <model_init> function to AdapterTrainer at its init without passing a model 

trainer = AdapterTrainer(
      model_init=model_init,
     # model=model,  # THIS SHOULD NOT BE REQUIRED IF <model_init> is provided
     args=training_arguments,
     train_dataset=train_dataset,
     eval_dataset=eval_dataset,
)

Error that I get:
AttributeError: 'NoneType' has no attribute 'active adapters'

The same code would work with standard HF transformers, with adapter-transformers it does not work.

I know that there is a workaround - just passing a mockup model that will never be used during hyperparameter search but this is not how the Library should behave. Maybe a simple if statement (if self.model and not self.model_init) before checking model properties is enought but maybe deeper check is required.

## Expected behavior

The same as with original HF Trainer - it should accept <model> OR <model_init> at the init phase

bartoszpawlowicz:
any idea when the official release with this bug fix will be published? I see thats it is already merged to master but last release is from September...

calpt:
@bartoszpawlowicz We're currently in the final preparations for the next release. Should be available some time in the next week.

PR #482, opened 2023-01-19
Fix adaptertrainer hyperparameter tuning

dtuit:
Prevents an error when calling `Trainer.hyperparameter_search`, which requires that Trainer is initialized with `model` set to `None` and `model_init` is used instead. Resolved by using `self.model`, since `super().__init__` will set the value of `self.model` using `model_init`.


calpt (approved):
Looks good! Thanks for fixing and providing a test case!

I believe this also resolves #481.

PR #487, opened 2023-01-28
Move loading best adapter to the trainer class

MaBeHen:
When activating ```load_best_model_at_end```, AdapterTrainer first tries (unsuccessfully) to load the full model.
Moving the adapter loading code to ```_load_best_model()``` and thereby overwriting it should fix this.


calpt (approved):
Looks good, thanks!

PR #488, opened 2023-01-30
Add Adapters to ALBERT

lenglaender:
This PR adds adapters to the ALBERT model.

All tests run through except some class conversation tests. I think the error must be in the `STATIC_TO_FLEX_HEAD_MAP` in `head_utils.py`, but I am not 100% sure. Can you take a look at this, @calpt or @hSterz ?


src/transformers/adapters/mixins/albert.py
```diff
@@ -0,0 +1,44 @@
...
+        self.attention_adapters._init_adapter_modules()
+
+
+class AlbertModelAdaptersMixin(EmbeddingAdaptersMixin, InvertibleAdaptersMixin, ModelAdaptersMixin):
+    """Adds adapters to the AlbertModel module."""
+
+    def iter_layers(self) -> Iterable[Tuple[int, nn.Module]]:
+        for i, albertLayerGroup in enumerate(self.encoder.albert_layer_groups):
+            for albertLayer in albertLayerGroup.albert_layers:
+                yield i, albertLayer
```
calpt, line 29: I'm not sure about the exact logic of Albert's layer sharing, but couldn't treating the layer group id here as layer id for each adapter be problematic for certain config keys (e.g. reduction factor per layer, leave_out)?

tests_adapters/test_adapter_embeddings.py
```diff
@@ -21,34 +21,34 @@ def test_load_embeddings(self):
 
         self.assertEqual(model.active_embeddings, "test")
 
-    def test_add_embeddings(self):
+    def test_add_embeddings(self, embedding_dim=None):
         model = self.get_model()
         tokenizer = AutoTokenizer.from_pretrained("tests_adapters/fixtures/SiBERT")
-        model.add_embeddings("test", tokenizer)
+        model.add_embeddings("test", tokenizer, embedding_dim=embedding_dim)
```
calpt, line 27: Here and other places with this change: Shouldn't we rather try to figure out `embedding_dim` automatically in `add_embeddings()` as `embedding_size` is part of the model config?

calpt (commented):
Looks pretty good to me overall 👍

src/transformers/adapters/mixins/albert.py
lenglaender: You are right - should be fixed by acefa59c3b3ebbf65eb92f140ca0fed492e4ed18

lenglaender:
Thanks for the review @calpt. Everything you noticed should be fixed now

src/transformers/adapters/mixins/albert.py
calpt: Thanks for fixing.

I think we should additionally add a note in the docs to clarify how the adapter implementation in Albert is different from others (i.e. that layers + adapters are shared and using layer ids in the adapter config will not affect the actual Albert layers, but the hidden layer groups + inner groups). I did a similar thing for CLIP here: https://raw.githubusercontent.com/adapter-hub/adapter-transformers/master/adapter_docs/classes/models/clip.rst

src/transformers/adapters/model_mixin.py
```diff
@@ -188,7 +193,7 @@ def load_embeddings(self, path: str, name: str):
         self.set_active_embeddings(name)
         return tokenizer
 
-    def add_embeddings(self, name, tokenizer, reference_embedding=None, reference_tokenizer=None, embedding_dim=None):
```
calpt, line 191: I'm not sure if the `embedding_dim` parameter can be removed without issues (as it was available before this PR). If we want to remove it, also remove the method docs of this param below.

calpt (approved):
Feel free to squash & merge once remaining two comments are resolved.

src/transformers/adapters/model_mixin.py
lenglaender: You're right, I added the parameter again

src/transformers/adapters/mixins/albert.py
lenglaender: Added a note to the docs in 4d2fb06 :+1:

Issue #509, opened 2023-03-06
Python version


fzalkow:
Hi,

The readme file specifies "Python 3.7+" as a requirement, which I understand as ">=3.7". However, the assignment operator `:=` is used in the file [`adapters/utils.py`](https://github.com/adapter-hub/adapter-transformers/blob/master/src/transformers/adapters/utils.py#L411), which is not allowed until Python 3.8.

Thanks for the great work with adapter-transformers!

lenglaender:
@fzalkow from now on we will officially only support python3.8+ as python3.7 is only supported until June this year. Nevertheless, the master branch should currently be fully functional with python3.7, but we cannot say how long this will be the case.

Thanks for letting us know that adapter-transformers no longer worked with Python 3.7!

PR #510, opened 2023-03-07
Fix python3.7 Compatibility

lenglaender:
Compatibility with Python 3.7+ an Pytorch 1.12.0+:
- Changed [`src/transformers/adapters/utils.py`](https://github.com/adapter-hub/adapter-transformers/blob/master/src/transformers/adapters/utils.py#L411) so that it remains compatible
- Adapted the python and pytorch versions in the CI/CD pipeline workflows


lenglaender:
Fixes #509

hSterz:
This looks good. But we should consider, whether we want to continue to support python 3.7. The adapter-transformers version 3.2 we released does not. So "going back" in the supported python version might be confusing. Additionally, python 3.7 is only supported till June this year. It might make more sense to update our documentation and only support version 3.8 or higher.

hSterz (approved):

Issue #506, opened 2023-03-02
`resume_from_checkpoint` errors `local variable 'adapter_loaded' referenced before assignment`


NightMachinery:
## Environment info

     
- `adapter-transformers` version: 3.2.0a0
- Platform: Linux-5.10.147+-x86_64-with-glibc2.29
- Python version: 3.8.10
- Huggingface_hub version: 0.12.1
- PyTorch version (GPU?): 1.13.1+cu116 (True)
- Tensorflow version (GPU?): 2.11.0 (True)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Using GPU in script?: yes
- Using distributed or parallel set-up in script?:

## Information

Model I am using (Bert, XLNet ...):
Bert

Language I am using the model on (English, Chinese ...):
Persian

Adapter setup I am using (if any):

The problem arises when using:
* [x] the official example scripts: (give details below)
* [ ] my own modified scripts: (give details below)

The tasks I am working on is:
* [ ] an official GLUE/SQUaD task: (give the name)
* [x] my own task or dataset: (give details below)

## To reproduce

The whole notebook can be found [here](https://colab.research.google.com/drive/1It9qONGo5sKHokfwoDs8Jo4Hoxt8FIB9?usp=sharing).

```python
! python run_mlm.py \
    --model_name_or_path bert-base-uncased \
    --train_file './digi_text_train.csv' \
    --validation_file './digi_text_test.csv' \
    --do_train \
    --do_eval \
    --learning_rate 1e-4 \
    --num_train_epochs 10.0 \
    --output_dir ./test-mlm \
    --train_adapter \
    --adapter_config "pfeiffer+inv" \
    --save_strategy steps \
    --evaluation_strategy steps \
    --load_best_model_at_end \
    --push_to_hub \
    --hub_model_id 'lang_adapter_fa_digikala' \
    --hub_strategy 'checkpoint' \
    --resume_from_checkpoint 'last-checkpoint'
```

```python
Traceback (most recent call last):
  File "run_mlm.py", line 619, in <module>
    main()
  File "run_mlm.py", line 568, in main
    train_result = trainer.train(resume_from_checkpoint=checkpoint)
  File "/usr/local/lib/python3.8/dist-packages/transformers/trainer.py", line 1532, in train
    self._load_from_checkpoint(resume_from_checkpoint)
[...]

PR #514, opened 2023-03-10
Fix resume_from_checkpoint

hSterz:
If the `resume_from_checkoint`is wrongly indented which leads to errors as described in #506


lenglaender (approved):
Looks good :+1:

PR #518, opened 2023-03-14
Fix LoRA & (IA)³ implementation for Bart & MBart

calpt:
This PR fixes a critical issue in the LoRA & (IA)³ implementation of Bart & MBart, where LoRA & (IA)³ weights were not added to the intermediate and output linear layers of the model's decoder blocks.

I.e., adapter configs having `intermediate_lora=True` or `output_lora=True` are added incorrectly to (M)Bart models. For LoRA, this does not affect the default config, for (IA)³ it does (`intermediate_lora=True`).

To ensure correct addition of weights in the future, `get_adapter()` tests are updated to count the number of modules added per adapter.


hSterz (approved):
Looks good to me

Issue #521, opened 2023-03-17
Adapter support for GPTNeoX


ajesujoba:
I have implemented adapter for GPTNeoX following the instructions in the documentation. It passed all tests but during the training of the language adapter, it trained the prediction head too. Do you by chance have an idea why this is happening?

calpt:
Hey @ajesujoba, this sounds great, would be awesome to have GPTNeoX support integrated into the library, so feel free to do a PR!

Regarding your question on language adapter training, could you add some more context on what you observed and which behavior you expected (ideally with some code snippet). Thank you!

calpt:
Thanks for providing the additional context (and of course thanks for opening the PR!).

After looking into it a bit deeper, the cause for this behavior seems to be that GPT-NeoX does not tie the weights of the input and output projection layers. By default, adapter-transformers will _only_ freeze the weights of the base model, excluding the weights of any prediction head (as you usually want to fine-tune that with the adapter). Thus, for LM heads, freezing the output projection layer relies on the fact that most models supported so far share these weights with the input projection (which is part of the base model, therefore frozen).

To ensure the expected behavior also for GPT-NeoX, we'd probably need to freeze the output projection manually somewhere in the code. Maybe adding it to the `freeze_model()` method in the model mixin via `self.get_output_embeddings()` would work.

ajesujoba:
Hi @calpt, thanks for your feedback. I thought as much, I also noticed that they did not tie the weights of the input and output projection layers. 

Yes, I agree that freezing the prediction head somewhere else such as within `freeze_model()` would be the best option. I guess this would be done at your end right?

calpt:
You can directly integrate a fix for this into your PR with the model integration if you like. Otherwise, I could also add it independently.

ajesujoba:
Checking again it appears it is not plausible to have it fixed within freeze_model(). `self` within `model_mixin.py` is the `base_model` without any prediction head (because the embeddings are not tied). I may be wrong. So I guess a possible place to fix this would be within  `training.py`. If it is fine with you, you can add it independently so that I don't break a lot of things

PR #523, opened 2023-03-21
Adding adapter support for NeoX

ajesujoba:
Added adapter support for GPTNeoX  with tests. Although at the moment, while training the adapter module, it also trains the CLM head (which is not the expected situation). This I already raised [here](https://github.com/adapter-hub/adapter-transformers/issues/521).


src/transformers/__init__.py
```diff
@@ -2942,6 +2944,15 @@
...
     )
+    _import_structure["models.gpt_neox"].extend(
+        [
+            "TFGPTNeoXForCausalLM",
+            "TFGPTNeoXForQuestionAnswering",
+            "TFGPTNeoXForSequenceClassification",
+            "TFGPTNeoXModel",
+            "TFGPTNeoXPreTrainedModel",
+        ]
+    )
```
calpt, line 2955: Is the addition of these imports related to changes of this PR?

src/transformers/adapters/models/gpt_neox/adapter_model.py
```diff
@@ -0,0 +1,219 @@
...
+        num_labels=2,
+        layers=1,
+        activation_function="tanh",
+        overwrite_ok=False,
+        id2label=None,
+    ):
+        head = QuestionAnsweringHead(self, head_name, num_labels, layers, activation_function, id2label)
+        self.add_prediction_head(head, overwrite_ok)
+
+class GPTNeoXModelWithHeads(GPTNeoXAdapterModel):
```
calpt, line 188: As XModelWithHeads classes are deprecated, we wouldn't want to add this class for newly supported models.

src/transformers/models/gpt_neox/modeling_gpt_neox.py
```diff
@@ -95,7 +105,7 @@ def __init__(self, config):
             self.rotary_ndims, config.max_position_embeddings, base=config.rotary_emb_base
         )
         self.norm_factor = torch.sqrt(torch.tensor(self.head_size, dtype=torch.float32)).to(torch.get_default_dtype())
-        self.query_key_value = nn.Linear(config.hidden_size, 3 * config.hidden_size)
+        self.query_key_value = LoRALinear(config.hidden_size, 3 * config.hidden_size, "selfattn", config)
```
calpt, line 108: As GPT-NeoX groups all attention matrices together into one linear layers similar to GPT-2, this should probably use `LoRAMergedLinear` instead of `LoRALinear` if i'm not mistaken (see GPT-2 implementation for reference).

calpt (commented):
@ajesujoba Sorry for the delay in getting back to you. I've done an initial review of your changes and left some comments. Overall, this looks very promising, thanks again for your efforts.

Before this can be merged, please additionally make sure:
- all checks and tests in the CI pipeline pass
- all points in [our contributing guide](https://docs.adapterhub.ml/contributing/adding_adapters_to_a_model.html) are addressed
- in `tests_adapters/models`, a new module for GPT-NeoX is added (this is currently not correctly documented in the contributing guide, we'll update)

Additionally, I've provided a fix for the issue regarding output embeddings you've raised and we've discussed in #537.

src/transformers/__init__.py
ajesujoba: No, they are not. Should I remove them

src/transformers/adapters/models/gpt_neox/adapter_model.py
ajesujoba: Sure, we can remove them.

ajesujoba: @calpt , one more thing to point out. [GPTNeoX configuration](https://github.com/adapter-hub/adapter-transformers/blob/master/src/transformers/models/gpt_neox/configuration_gpt_neox.py) does not include `'hidden_dropout_prob' ` and `attention_probs_dropout_prob`. And running test_adapters with this would result in an error. What do you suggest?

src/transformers/__init__.py
ajesujoba: No, I removed them

ajesujoba: Regarding `all points in [our contributing guide](https://docs.adapterhub.ml/contributing/adding_adapters_to_a_model.html) are addressed`, it is optional in the documentation and I tried it but having some mismatches in dimension, Is it still optional?

calpt: Could you clarify what exactly you're referring to from the contributing guide? The Parallel inference and static head conversion points are still optional (although highly recommended). If Parallel support is not implemented, please make sure to remove the test mixin classes starting with "Parallel..." from the model test class.

src/transformers/adapters/models/gpt_neox/adapter_model.py
calpt: Regarding the tokenizer, I think we should we should allow using fast tokenizers anyway. Could you check if/ which issues occur when you switch to loading the fast tokenizer for GPTNeoX?

Regarding the dropout issue, we would probably add some default dropout if the keys are not present in the model config. I can look into this. You could just add the missing key to the model config temporarily to make the tests pass and we can change it later on.

src/transformers/__init__.py
ajesujoba: Yes, I was referring to the Parallel inference and static head conversion points are still optional (although highly recommended). And as you recommended I would remove test mixin classes starting with "Parallel...". I would also remove some of the other tests such as `IA3TestMixin, LoRATestMixin, PrefixTuningTestMixin, UniPELTTestMixin,` as they require adding classification head `'add_classification_head'`  and the GPT_NeoX model in this version does not have that .

src/transformers/adapters/models/gpt_neox/adapter_model.py
ajesujoba: Ok! Sounds good

src/transformers/models/gpt_neox/modeling_gpt_neox.py
ajesujoba: I fixed this already!

calpt:
Hey, thanks again for you efforts in contributing new model architectures to adapter-transformers and sorry for the silence on our side.

In the last few weeks, we've been working on a large refactoring of our project, which will ultimately result in the release of _Adapters_, the next-generation adapters library. See https://github.com/adapter-hub/adapter-transformers/issues/584.

As a consequence, we plan to merge any new model integrations directly to the new codebase, which currently can be found [on this branch](https://github.com/adapter-hub/adapter-transformers/tree/adapters). Unfortunately, this necessitates some changes in the model integration code (detailed [here](https://github.com/adapter-hub/adapter-transformers/blob/adapters/docs/contributing/adding_adapters_to_a_model.md), see already integrated models such as BERT, BART etc. for reference).

If you'd be willing to update your model integration to target the new library yourself, we'd be super happy to help you on this. Otherwise, we might look into upgrading and merging some of the open model integration PRs ourselves in the future. For more details, again see https://github.com/adapter-hub/adapter-transformers/issues/584.

PR #537, opened 2023-04-19
Make sure output embeddings are frozen when training adapters

calpt:
This PR:
- resolves the issue reported in #521 and described in https://github.com/adapter-hub/adapter-transformers/issues/521#issuecomment-1485709999
- fixes get_/set_output_embeddings() when having multiple flex heads


src/transformers/adapters/heads/base.py
```diff
@@ -509,17 +509,25 @@ def _init_head_modules(self):
...
+            return embeddings
 
     def set_output_embeddings(self, new_embeddings):
         # Only sets the output embeddings for the currently active head
-        if self.active_head in self.heads:
-            head = self.heads[self.active_head]
-            if head.get_output_embeddings() is not None:
+        for head_name in self._active_heads:
+            if head_name in self.heads:
+                head = self.heads[head_name]
```
hSterz, line 530: Do I understand it correctly that, in case f multiple heads, with the current implementation `get_output_embeddings()` returns a list of embeddings, and `set_output_embeddings()` only takes one embedding as input that will be set for every head? If so this is kind of unintuitiv

calpt: Yeah, the implementation is not ideal here. We need `get_output_embeddings()` to return the embeddings of all active heads to make sure we can access them eg. in `freeze_embeddings()`. What would your suggestion be to implement this better?

calpt: done :)

calpt:
Superseded by #600.

PR #543, opened 2023-05-06
Fix Notebook01 Dataset column_rename

julianpollmann:
Fix notebook dataset column_rename() issue.


calpt (approved):
Thanks for fixing this!

PR #581, opened 2023-08-28
Add adapter support for X-MOD model

calpt:
Integrates adapter support into HuggingFace's X-MOD implementation.

- X-MOD paper: http://dx.doi.org/10.18653/v1/2022.naacl-main.255
- HF docs: https://huggingface.co/docs/transformers/model_doc/xmod

### Implementation notes

HF model already supports adapters integrated into X-MOD. Several changes were necessary to make HF's implementation compatible with Adapters:
- deactivate HF's custom adapter logic: `lang_ids` parameter and `set_default_language()` are deactivated in favor of regular adapter methods. Documented in model class docs.
- conversion of adapter checkpoints: `utils/convert_xmod_checkpoint.py` Only converted checkpoints can be used with Adapters. We'll convert & re-upload some of the original X-MOD checkpoints to the Hub.

### Tasks
- [x] basic model implementation
- [x] tests
- [x] documentation
- [x] convert weights from HF checkpoints
- [x] upload converted checkpoints

### DIFF
--- a/docs/classes/models/xmod.rst
+++ b/docs/classes/models/xmod.rst
@@ -1,0 +1,23 @@
+X-MOD
+=====
+
+.. note::
+    The X-MOD implementation integrated into Transformers already supports adapters.
+    To make this implementation compatible with Adapters, a few changes were necessary:
+
+        - In Adapters, the X-MOD classes rely on the usual adapter methods instead of the custom methods introduced in Transformers, i.e.:
+            - ``set_active_adapters()`` instead of ``set_default_language()``.
+            - ``AdapterSetup`` context instead of ``lang_ids`` parameter.
+        - We provide dedicated model checkpoints converted for usage with Adapters
+            - e.g. ``facebook/xmod-base`` is available as ``AdapterHub/xmod-base`` with languages adapters split into separate repos (e.g. ``AdapterHub/xmod-base-af_ZA``) for on-demand loading.
+
+The abstract from the paper is the following:
+
+*Multilingual pre-trained models are known to suffer from the curse of multilinguality, which causes per-language performance to drop as they cover more languages. We address this issue by introducing language-specific modules, which allows us to grow the total capacity of the model, while keeping the total number of trainable parameters per language constant. In contrast with prior work that learns language-specific components post-hoc, we pre-train the modules of our Cross-lingual Modular (X-MOD) models from the start. Our experiments on natural language inference, named entity recognition and question answering show that our approach not only mitigates the negative interference between languages, but also enables positive transfer, resulting in improved monolingual and cross-lingual performance. Furthermore, our approach enables adding languages post-hoc with no measurable drop in performance, no longer limiting the model usage to the set of pre-trained languages.*
+
+XmodAdapterModel
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. autoclass:: adapters.XmodAdapterModel
+    :members:
+    :inherited-members: XmodPreTrainedModel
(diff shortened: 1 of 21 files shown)


src/adapters/models/xmod/mixin_xmod.py
```diff
@@ -0,0 +1,71 @@
...
+        return super().forward(*args, **kwargs)
+
+    # Override adapter-specific methods in original implementation
+
+    def set_default_language(self, language: str):
+        raise ValueError(
+            "`set_default_language` is not implemented for models using `adapters`. Use `set_active_adapters` instead."
+        )
+
+    def freeze_embeddings_and_language_adapters(self):
```
hSterz, line 60: Is this a function, that would be helpful for other models as well? If so, would it make sense to move it to the model_mixin?

src/adapters/models/xmod/modeling_xmod.py
```diff
@@ -0,0 +1,168 @@
...
+from torch import nn
+
+from transformers.models.xmod.modeling_xmod import XmodOutput, XmodSelfAttention, XmodSelfOutput
+
+from ...composition import adjust_tensors_for_parallel
+from ..bert.mixin_bert import BertOutputAdaptersMixin, BertSelfAttentionAdaptersMixin, BertSelfOutputAdaptersMixin
+
+
+# Copied from transformers.models.roberta.modeling_roberta.RobertaSelfAttention with Roberta->Xmod
+class XmodSelfAttentionWithAdapters(BertSelfAttentionAdaptersMixin, XmodSelfAttention):
```
hSterz, line 31: The copied classes are missing from the [table](https://www.notion.so/2a4ed3a4038041c495c4927a323d0e2b?v=0efccebe90634b40816a7886aae806ee)

src/adapters/models/xmod/modeling_xmod.py
```diff
@@ -0,0 +1,168 @@
...
+    def forward(self, hidden_states: torch.Tensor, input_tensor: torch.Tensor) -> torch.Tensor:
+        hidden_states = self.dense(hidden_states)
+        hidden_states = self.dropout(hidden_states)
+        hidden_states = self.adapter_layer_forward(hidden_states, input_tensor, None)
+        return hidden_states
+
+
+class XmodOutputWithAdapters(BertOutputAdaptersMixin, XmodOutput):
+    def __init__(self, config):
+        super().__init__()
```
hSterz, line 149: Since this class is normally dynamically stitched in during runtime. In which case is the init method called?

docs/classes/models/xmod.rst
```diff
@@ -0,0 +1,23 @@
...
+=====
+
+.. note::
+    The X-MOD implementation integrated into Transformers already supports adapters.
+    To make this implementation compatible with Adapters, a few changes were necessary:
+
+        - In Adapters, the X-MOD classes rely on the usual adapter methods instead of the custom methods introduced in Transformers, i.e.:
+            - ``set_active_adapters()`` instead of ``set_default_language()``.
+            - ``AdapterSetup`` context instead of ``lang_ids`` parameter.
+        - We provide dedicated model checkpoints converted for usage with Adapters
```
hSterz, line 11: I think this should make it more clear that the original checkpoints do not work with the `adapters` model.

hSterz (commented):
Looks good just some minor comments

src/adapters/models/xmod/mixin_xmod.py
calpt: This method is built-in by HF for X-Mod, so it's currently overriden in the mixin to make it compatible with adapters. We can think about making it available to all models in the future.

docs/classes/models/xmod.rst
calpt: Tried to make it a bit clearer.

src/adapters/models/xmod/mixin_xmod.py
hSterz: Ah, I see :) That sounds like a plan

hSterz (approved):
Looks good

PR #582, opened 2023-08-31
Update uploading in PushAdapterToHubMixin

calpt:


hSterz (approved):

src/adapters/hub_mixin.py
```diff
@@ -184,7 +183,7 @@ def push_adapter_to_hub(
         use_temp_dir = not os.path.isdir(local_path) if local_path else True
 
         # Create repo or get retrieve an existing repo
-        repo_id, token = self._create_repo(repo_id, private=private, use_auth_token=use_auth_token)
```
TimoImhof, line 187: Is `token` the same as `use_auth_token`?

calpt: yes, HF also uses it like this

PR #583, opened 2023-09-04
Electra Support

hSterz:
Adding support for the ElectraModel. Since the ElectraModel has several classes copied from the BertModel, the mixins from the BertModel are reused.


calpt (approved):
Looks good, thanks for working on it!

PR #588, opened 2023-09-25
[WIP] Add MLM-based audio models (HuBERT, WavLM, and MERT) support

yamathcy:
Hello, thank you for organizing and refactoring the awesome library.
I added HuBERT(https://arxiv.org/abs/2106.07447) and WavLM (https://arxiv.org/abs/2110.13900).
I also modified MERT (https://arxiv.org/abs/2306.00107) model, which is a variation of HuBERT for musical audio. (its mixin is borrowed from mixin of HuBERT)


yamathcy:
I'm sorry for the wrong pull request...

PR #594, opened 2023-10-17
Automatically convert heads when loading with XAdapterModel

### LINKED ISSUE #592: How to load the original model LM head in the new version of Adapters?
## Details

Hello!

We are working with LLaMA which is not supported in the old version of Adapter-Transformers. So we decided to use the beta (`Adapters` branch). The issue is, it's unclear how to load the llama language modeling head. By default, the model is loaded without any head, which is a problem. This kind of approach made a lot of sense in the BERT era, but it's unclear if it's an intuitive solution now when people want to fine-tune language models with their original head on.

Here's what we tried:
```python
from adapters import LlamaAdapterModel
from transformers import AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-2-7b-hf")
model = LlamaAdapterModel.from_pretrained("meta-llama/Llama-2-7b-hf")

model.add_causal_lm_head("lm_head")
adapters.init(model)
out = model.generate(tokenizer("The cat is", return_tensors="pt")["input_ids"], max_length=20)
tokenizer.decode(out[0])
# output: '<s> The cat isРСРoulevalu str siècleTest materadejkradeavenradebosebose étudesvä'
```

so, it seems like this way the added head is randomly initialized.

Could you help us to get the original llama head weights?

calpt:
Related to #592.
Changes loading logic of XAdapterModel classes such that static heads of loaded checkpoint will be automatically converted.
E.g.

```python
model = AutoAdapterModel.from_pretrained("bert-base-uncased")

print(model.heads)
```
should automatically create a masked-LM head using the weights of the original head from the pre-trained checkpoint.


calpt:
There's currently still an open issue for models such as BERT, where the LM head bias is not correctly:

```python
from adapters import AutoAdapterModel

model, loading_info = AutoAdapterModel.from_pretrained("bert-base-uncased", output_loading_info=True)
print(model.heads)
print(loading_info)
```

The cause of this issue is unrelated to the current changes.

lenglaender (approved):
Looks solid to me :+1:

Issue #584, opened 2023-09-04
Developing the next-generation adapters library


calpt:
As you might have noticed in the past weeks and months as a user of AdapterHub/ adapter-transformers, updates from our side (and replies to opened issues) have been quite scarce. This is because we've been busy working on a major revamp of our library, which will be officially released in the coming weeks. This post shall give you a preview on what's coming.

## What's changing?

- We've fundamentally refactored our code and they way adapters are integrated into Transformer models. Most importantly, this means **the new library will no longer be a fork of HuggingFace's Transformers, but a separate add-on package.** We hope to alleviate all sorts of issues related to the previous forking strategy with this change.
- New model integrations, adapter methods and more...
- A new name: To highlight the fundamental library design changes, **we're giving the new library a new name: _Adapters_**. Thus, the `adapters` package will replace the `adapter-transformers` package in the future.

For a detailed breakdown on all breaking changes, check out https://github.com/adapter-hub/adapter-transformers/blob/adapters/docs/transitioning.md.

## Where can I test it?

The current development version of _Adapters_ is available here: https://github.com/adapter-hub/adapter-transformers/tree/adapters. We're in the process of adding more features before releasing it in the main branch of this repo in the coming weeks.

## How can I help?

We've updated our [contributing guide](https://github.com/adapter-hub/adapter-transformers/blob/adapters/CONTRIBUTING.md) as well as the guides for [adding new model integrations](https://github.com/adapter-hub/adapter-transformers/blob/adapters/docs/contributing/adding_adapters_to_a_model.md) and [adding new adapter methods](https://github.com/adapter-hub/adapter-transformers/blob/adapters/docs/contributing/adding_adapter_methods.md) on the development branch. We're always happy for your help!

## What are the next steps?

- Finalizing all new features of the new library and fixing remaining bugs
- _Adapters_ will be officially released in this repo and as package on PyPI, replacing the current adapter-transformers library (keeping the old library in an archived version)
- In parallel, we'll try to catch up on open issues and bug reports. Thanks for all your patience in this regard!

---

Please feel free to comment any of your thoughts on all of this below :)

Current team: @hSterz, @lenglaender, @TimoImhof, @calpt

cody-moveworks:
Hi! Is there a rough ETA for when the new `adapters` package will be released? Very excited to use this new package :)

hSterz:
Hi @cody-moveworks, we plan to release the `adapters` package within the next 2 weeks.

[...]

PR #595, opened 2023-10-18
Add Prompt Tuning

lenglaender:
Depends on: #591 

This PR adds Prompt Tuning (https://aclanthology.org/2021.emnlp-main.243/)

Currently the Prompt Tuning Layer was created and the BERT model has been adapted


calpt (commented):
did a first review pass, looks good overall, haven't checked all details though

calpt:
Please also add row in table on main readme

raphaelreimann:
Hey, love the integration of prompt tuning! I would love to use it when it is stable, do you know when this will be merged? 🙂 @lenglaender @calpt

lenglaender:
@raphaelreimann We're currently fixing the last few tests that aren't running. For every model except CLIP, EncoderDecoder, GPT-2, GPT-J, LLaMA and T5 the implementation is already working.

As soon as this PR and the sync PR (#602) are merged, we will release the new Adapters library (#584).

raphaelreimann:
@lenglaender Thank you for the update on this, great work!

src/adapters/model_mixin.py
```diff
@@ -40,22 +41,6 @@ def init_adapters(self, model_config, adapters_config, **kwargs):
         if hasattr(super(), "init_adapters"):
             super().init_adapters(self.config, self.adapters_config, **kwargs)
 
-        self.hook_after_embeddings(self._hook_fn)
-
-    def _hook_fn(self, module, args, output):
-        new_output = self.invertible_adapters_forward(output)
-        return new_output
-
-    def hook_after_embeddings(self, hook_fn: Callable):
```
hSterz, line 49: Why are we deleting/refactoring this?

tests_adapters/methods/test_prompt_tuning.py
```diff
@@ -0,0 +1,42 @@
...
+        model = self.get_model()
+        self.run_forward_test(model, PromptTuningConfig(prompt_length=10))
+
+    def test_load_prompt_tuning(self):
+        self.run_load_test(PromptTuningConfig(prompt_length=10))
+
+    def test_load_full_model_prefix_tuning(self):
+        self.run_full_model_load_test(PromptTuningConfig(prompt_length=10))
+
+    def test_train_prefix_tuning(self):
```
hSterz, line 41: Probably, a left over from copying over tests but I guess this should be called `test_train_prompt_tuning`

hSterz (approved):
Looks good to me just a few minor comments

src/adapters/model_mixin.py
lenglaender: We now have 2 adapter methods that are hooked in after the embeddings (invertible adapter & prompt tuning) and have therefore decided to put them in the new `post_embedding_forward` function of the `ModelBaseAdaptersMixin`. This makes the implementation cleaner and it is easier to add additional adapter methods later, which need to be added after the embedding layer as well.

tests_adapters/methods/test_prompt_tuning.py
lenglaender: Yes, thanks for pointing that out! I've renamed it

lenglaender:
For the subsequent PR, which will extend Prompt Tuning with composition support and which will add support for the not yet supported models, we should think about realising the change of the attention mask for prefix tuning within the function `prefix_attention_mask`.

PR #598, opened 2023-11-07
Add Composition Support to LoRA and (IA)³

calpt:
Follow-up to #591.

This PR provides initial support for adapter composition in LoRA & (IA)³ modules. Currently LoRA & (IA)³ don't support composition. With this PR, the following blocks will be supported: **Stack, BatchSplit, Average, Parallel**

Additionally, the LoRA implementation is refactored a bit in an effort to make it cleaner.

### Limitations
- Split & Fuse compositions are **not** supported
- LoRA/ (IA)³ composition is **not** supported for models using the `LoRAMergedLinear` implementation. These currently are: **GPT-2, DeBERTa (v1)**

### TODO
- [x] Fix remaining errors
- [x] Documentation

### DIFF
--- a/src/adapters/methods/adapter_layer_base.py
+++ b/src/adapters/methods/adapter_layer_base.py
@@ -151,7 +151,7 @@ class ComposableAdapterLayerBase(AdapterLayerBase):
 
     Make sure the 'adapter_modules_name' and 'supported_compositions' attributes as well as all abstract methods are
     overriden in derived classes. 'allow_multi_parallelize' can be set to True to allow inputs to be parallelized
-    independently multiple times. This is useful when there are multiple parallel input flows though an adapter layer
+    independently multiple times. This is useful when there are multiple parallel input flows through an adapter layer
     (e.g. in LoRA).
     """
 
--- a/src/adapters/methods/adapter_layer_base.py
+++ b/src/adapters/methods/adapter_layer_base.py
@@ -154,6 +154,7 @@ class ComposableAdapterLayerBase(AdapterLayerBase):
     """
 
     supported_compositions = []
+    allow_multi_parallelize = False
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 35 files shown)


TimoImhof:
I already did a run with the backwards compatibility tests and they pass ✅

src/adapters/methods/lora.py
```diff
@@ -39,58 +40,121 @@ def __init__(
...
+        self.scaling = self.lora_alpha / self.r
+
+        if config.init_weights == "lora":
+            # initialize A the same way as the default for nn.Linear and B to zero
+            nn.init.kaiming_uniform_(self.lora_A, a=math.sqrt(5))
+            nn.init.zeros_(self.lora_B)
+        elif config.init_weights == "bert":
+            nn.init.normal_(self.lora_A, std=0.02)
+            nn.init.normal_(self.lora_B, std=0.02)
+        elif config.init_weights == "ia3":
```
hSterz, line 54: I am not super familiar with the initialization methods, but is it intended that lora still has ia3 initialization and the other way around?

src/adapters/methods/lora.py
```diff
@@ -39,58 +40,121 @@ def __init__(
...
+
+        return hidden_states, gate
+
+
+class IA3(nn.Module):
+    def __init__(
+        self,
+        lora_A_shape,
+        lora_B_shape,
+        config: LoRAConfig,
```
hSterz, line 97: Why is this not IA3 config? With separating the modules it might make sense to consider also having separate config classes

src/adapters/methods/lora.py
```diff
@@ -39,58 +40,121 @@ def __init__(
...
+        self.lora_alpha = config.alpha
+        self.composition_mode = config.composition_mode
+        self.attn_matrices = config.attn_matrices
+        self.use_gating = config.use_gating
+        # Optional dropout
+        if config.dropout > 0.0:
+            raise ValueError("IA3 module does not support dropout.")
+
+        # Actual trainable parameters
+        self.lora_B = nn.Parameter(torch.zeros(lora_B_shape))
```
hSterz, line 114: Why are the attributes still named as `lora_alpha`, `lora_B` In the context of IA3?

src/adapters/methods/adapter_layer_base.py
```diff
@@ -154,6 +154,7 @@ class ComposableAdapterLayerBase(AdapterLayerBase):
     """
 
     supported_compositions = []
+    allow_multi_parallelize = False
```
hSterz, line 157: What is this parameter for? Maybe add short documentation?

hSterz (commented):
Looks good. I have just a few comments

src/adapters/methods/lora.py
calpt: yes, to avoid throwing an error. Using these combinations might not make sense though.

src/adapters/methods/lora.py
calpt: Separating configs is difficult because changing any config attribute would be a breaking change.

src/adapters/methods/lora.py
calpt: To be backwards compatible to older checkpoints.

TimoImhof (approved):
It looks good to me; thanks for working on this! I only left some minor comments.

src/adapters/methods/lora.py
TimoImhof: I think we should include this with a small comment for clarity in the code.

PR #599, opened 2023-11-07
Fix training T5 adapter models with Trainer

calpt:
Fixes https://www.notion.so/Fix-T5-f45601d8f58a49f3a2764d83a0ac645e?pvs=4


TimoImhof (approved):
LGTM, also tested both scripts again locally and they run ✅

PR #602, opened 2023-11-14
Sync/v4.35.0

lenglaender:


setup.py
```diff
@@ -61,7 +61,7 @@
     "sphinx-multiversion",
     "timeout-decorator",
     "torch>=1.10,!=1.12.0",
-    "transformers==4.33.3",
+    "transformers==4.35.0",
```
calpt, line 64: Can we update to the latest HF release, which is v4.35.2?

tests/tests_adapters/test_adapter_backward_compability.py
```diff
@@ -0,0 +1,48 @@
+import json
```
calpt, line 1: Why is this module in a separate "test_adapters" folder? Also, why is it marked as new instead of renamed/ moved?

calpt (commented):
Looks good overall, just a few questions left. Thanks for refactoring the test folder structure!

tests/tests_adapters/test_adapter_backward_compability.py
lenglaender: This file was unintentionally duplicated in this folder. I deleted it.

setup.py
lenglaender: I updated it to v4.35.2 now :+1:

PR #609, opened 2023-11-26
Add mistral support 

KorventennFR:
This PR is the follow up to #605 , adding adapter support for MistralAI's Mistral model.
It should fix the previous failed tests.


tests/test_mistral.py
```diff
@@ -0,0 +1,64 @@
...
+    ParallelAdapterInferenceTestMixin,
+    ParallelTrainingMixin,
+    MistralAdapterTestBase,
+    unittest.TestCase,
+):
+    pass
+
+
+@require_torch
+class BertClassConversionTest(
```
calpt, line 59:
```suggestion
class MistralClassConversionTest(
```

calpt (commented):
Looks very good overall, just one minor comment. Please also make sure to pass the style checks before we can merge it. 

Thanks a lot for your work on this!

KorventennFR:
Thank you for taking the time to review this. I had to code on a windows machine where I couldn't use or install Make but I will be able to do that very soon. I also noticed an unexpected behavior when doing inference after loading Mistral models with the adapters library. I will look into that before pushing again.

calpt (approved):

PR #612, opened 2023-12-02
Add loader methods for loading from/ to state_dict

calpt:
This PR adds methods for loading from/ storing to state_dicts in AdapterLoader.


TimoImhof (approved):
The changes look good to me!
Could you additionally add a small description of the PR for keeping track of when we changed what and why?

PR #621, opened 2023-12-18
Fix Prefix-Tuning for T5 models where d_kv != d_model / num_heads

### LINKED ISSUE #485: Prefix Tuning does not work with T5-3B
## Environment info

     
- `adapter-transformers` version: 3.1.0

## Information

Model I am using: T5-3B

Adapter setup I am using (if any): Prefix Tuning

## To reproduce

I believe this bug occurs every time T5-3B and Prefix Tuning are used, for example, it can be reproduced by running the summarization example:

```bash
python examples/pytorch/summarization/run_summarization.py \
    --model_name_or_path t5-small \
    --do_train \
    --do_eval \
    --dataset_name cnn_dailymail \
    --dataset_config "3.0.0" \
    --source_prefix "summarize: " \
    --output_dir /tmp/tst-summarization \
    --per_device_train_batch_size=4 \
    --per_device_eval_batch_size=4 \
    --overwrite_output_dir \
    --predict_with_generate \
    --train_adapter \
    --adapter_config prefix_tuning
``` 

Error is raised when prefix embeddings should be concatenated with keys and values:

```bash
  File "/home/hpcsapa1/.conda/envs/adapters-torch1.9/lib/python3.9/site-packages/transformers/adapters/prefix_tuning.py", line 329, in forward
    key_states = torch.cat([prefix_keys, key_states], dim=2)
RuntimeError: Sizes of tensors must match except in dimension 2. Expected size 32 but got size 128 for tensor number 1 in the list.
``` 

## Expected behavior

There is a mismatch in the sizes of prefix embeddings and keys and values. I think this is because the prefix size is based on [```hidden_size```](https://github.com/adapter-hub/adapter-transformers/blob/master/src/transformers/adapters/prefix_tuning.py#L171) which is defined as [```d_model```](https://github.com/adapter-hub/adapter-transformers/blob/master/src/transformers/adapters/wrappers/configuration.py#L35) for T5 models while it should be ```num_heads * d_kv```. For T5-Small, T5-Base and T5-Large ```d_model == num_heads * d_kv``` (and these models work fine) but this is not true for T5-3B where ```d_model=1024``` and ```num_heads=32```, ```d_kv=128``` .

calpt:
Fixes #485.

Allows passing head dim (`n_embd_per_head`) explicitly to prefix tuning to accomodate models where head dim is not equal to hidden dim / n_heads.

### DIFF
--- a/src/adapters/methods/prefix_tuning.py
+++ b/src/adapters/methods/prefix_tuning.py
@@ -21,19 +21,20 @@ class PrefixTuning(nn.Module, ModuleUtilsMixin):
         n_heads: int,
         input_size: int,
         config: PrefixTuningConfig,
+        n_embd_per_head: Optional[int] = None,
     ):
         super().__init__()
         self.n_layers = n_layers
         self.n_heads = n_heads
         self.input_size = input_size
-        self.n_embd_per_head = self.input_size // self.n_heads
+        self.n_embd_per_head = n_embd_per_head or self.input_size // self.n_heads
         self.config = config
 
         self.wte = nn.Embedding(self.config.prefix_length, self.input_size)
         self.control_trans = nn.Sequential(
             nn.Linear(self.input_size, self.config.bottleneck_size),
             Activation_Function_Class(self.config.non_linearity.lower()),
-            nn.Linear(self.config.bottleneck_size, self.n_layers * 2 * self.input_size),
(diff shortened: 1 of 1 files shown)


hSterz (approved):
Looks good to me :)

PR #624, opened 2023-12-21
[Bart] Move CLS rep extraction from EOS tokens to head classes

### LINKED ISSUE #494: Weird condition for sequence classification in Bart
## Environment info
- `adapter-transformers` version: 3.1.0

## Information
Model I am using (Bert, XLNet ...): Bart
I think checking condition for sequence classification by shape is quite buggy.
I wanted to use LM head with same enc & dec seq_len (and, of course, I could have different number of eos tokens), but conditional code below make it difficult to do that.
I work around with setting dec_seq_len == enc_seq_len-1, but it should be fixed for potential bug.
https://github.com/adapter-hub/adapter-transformers/blob/master/src/transformers/adapters/models/bart/adapter_model.py#L96

### LINKED ISSUE #563: `ValueError` warning in BartAdapterModel is unnecessary for sequence labeling task/generative task
## Information

Model I am using (Bert, XLNet ...): BartAdapterModel

Language I am using the model on (English, Chinese ...): English

Adapter setup I am using (if any): pffeifer

The problem arises when using:
* [ ] the official example scripts: (give details below)

The tasks I am working on is:
* [ ] an official GLUE/SQUaD task: mlm

## To reproduce

1. Follow the official [run_mlm.py](https://github.com/adapter-hub/adapter-transformers/blob/master/examples/pytorch/language-modeling/run_mlm.py) and use **BartAdapterModel** as the backbone.
2. Got " ValueError("All examples must have the same number of <eos> tokens.")" during training.

I checked the source code in [bart/adatper_model.py](https://github.com/adapter-hub/adapter-transformers/blob/accf70fa8b5fb6219c3ac0f8e2f826412425fce6/src/transformers/adapters/models/bart/adapter_model.py) and [modeling_bart.py](https://huggingface.co/transformers/v4.11.3/_modules/transformers/models/bart/modeling_bart.html) (in transformers) and found this in `BartAdapterModel`'s `forward()` function: 
```  
            if len(torch.unique(eos_mask.sum(1))) > 1:
                raise ValueError("All examples must have the same number of <eos> tokens.")
            cls_representation = x[eos_mask, :].view(x.size(0), -1, x.size(-1))[:, -1, :]

```
which is the same with `BartForSequenceClassification`. I think it's not necessary for a sequence labeling task? When doing masked language model, all the input text are concatenated and then grouped by every 1024 tokens, and it's common to include different number of <eos> tokens accordingly. 
This error exists even I add a `seq2seq_lm_head`.

## Expected behavior

The mentioned `ValueError` should be applied only for classification tasks.

calpt:
Fixes #494 and fixes #563.

`BartAdapterModel` currently tries to extract CLS token representations from EOS tokens independent from the used heads (if at all). This causes weird issues for models not using CLS token based heads.

This PR moves CLS token rep extraction into the specific head classes to only compute it when needed. Two (re-usable) new args to `forward_head()` are introduced:
- `get_cls_from_eos_tokens`: to indicate a model uses CLS reps from EOS tokens
- `eos_mask`: token mask to extract CLS reps (required when passing `get_cls_from_eos_tokens=True`)

### DIFF
--- a/src/adapters/heads/base.py
+++ b/src/adapters/heads/base.py
@@ -18,7 +18,13 @@ from transformers.modeling_outputs import (
 )
 from transformers.utils import ModelOutput
 
-from ..composition import AdapterCompositionBlock, BatchSplit, Parallel, parse_heads_from_composition
+from ..composition import (
+    AdapterCompositionBlock,
+    BatchSplit,
+    Parallel,
+    adjust_tensors_for_parallel,
+    parse_heads_from_composition,
+)
 from ..context import AdapterSetup, ForwardContext
 from ..loading import PredictionHeadLoader
 from ..methods.modeling import Activation_Function_Class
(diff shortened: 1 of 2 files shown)


lenglaender (approved):
Looks good :+1:

PR #625, opened 2023-12-23
readme: minor fix for installation section

stefan-it:
Hi :)

this PR has  a minor fix for the installation (from source) section: the `git checkout` command is removed, because there's no longer a `adapters` branch here.


calpt (approved):
Good catch, thank you!

PR #642, opened 2024-01-31
Fix obvious typo in docs

aranjan25:
Exactly what the title says -- removed duplicated text from the overview page in the docs.


calpt (approved):
thanks for fixing!

PR #641, opened 2024-01-28
Simplify XAdapterModel implementations

calpt:
This PR includes some refactorings of all AdapterModel classes:
- moves prediction head adding methods to base mixin & remove from all model classes
- convert `head_types` to list, use single mapping in base mixin

PR #654, opened 2024-02-23
Upgrade Transformers to v4.38.x

lenglaender:
Changes:
- HF changed parts of the Llama model implementation
- HF added a `LlamaForQuestionAnswering`. However, this model has a wrong base model name. I added a workaround that solves this problem until this is fixed in Transformers  (https://github.com/huggingface/transformers/pull/29258)


src/adapters/models/llama/adapter_model.py
```diff
@@ -149,3 +152,14 @@ def prepare_inputs_for_generation(
             }
         )
         return model_inputs
+
+    def _load_pretrained_model(cls, model, state_dict, loaded_keys, *args, **kwargs):
```
calpt, line 156: Why is this rewriting of the state dict required for us but not for HF?

lenglaender: I believe this is the case because when using Hugging Face, the question answering llama is loaded using the `AutoModelForQuestionAnswering` class, which utilises the models in `MODEL_FOR_QUESTION_ANSWERING_MAPPING_NAMES` and this contains `LlamaForQuestionAnswering`. `LlamaForQuestionAnswering` has "transformer" as the base model (https://github.com/huggingface/transformers/blob/2209b7afa04b3a6366350065f541e9248d6663c2/src/transformers/models/llama/modeling_llama.py#L1460)

But we load all models using the same `AutoAdapterModel` class: In `LlamaAdapterModel`, we set `self.model = LlamaModel(config)` and because of this the `state_dict` and `model` keys must have "model" as the base model.

lenglaender:
Removed `_load_pretrained_model` for the `LlamaAdapterModel` and updated the docs.

While updating the docs, I noticed that our sphinx version had some issues that others encountered, too: https://github.com/sphinx-doc/sphinx/issues/11890
Hence, I set the version to 5.0.2. This seems to change nothing; at least, I couldn't find any visual changes in the new build docs.

I also noticed that since #641 the docs don't include the add_XXX_head methods for every model, we have to fix this. I didn't find a quick fix, so I suppose we have to fix this in a separate PR.

@calpt I believe we can merge this PR now after you approve it.

docs/classes/models/auto.rst
```diff
@@ -4,6 +4,12 @@ Auto Classes
 Similar to the ``AutoModel`` classes built-in into HuggingFace Transformers, adapters provides an ``AutoAdapterModel`` class.
 As with other auto classes, the correct adapter model class is automatically instantiated based on the pre-trained model passed to the ``from_pretrained()`` method.
 
+```{eval-rst}
+.. note::
+    If the model loaded with the ``from_pretrained(...)`` function has a head, this head gets loaded as well. However, this only works for non-sharded models. If you want to load a sharded model with a head, you first need to load the model and then the head separately.
+    However, only LLMs are sharded. LLMs are typically used without a head, so this limitation of the ``from_pretrained(...)`` function should rarely occur.
```
calpt, line 10:
```suggestion
```

calpt:
> I also noticed that since https://github.com/adapter-hub/adapters/pull/641 the docs don't include the add_XXX_head methods for every model, we have to fix this. I didn't find a quick fix, so I suppose we have to fix this in a separate PR.

In the live version of our docs, we do have those, right? E.g. https://docs.adapterhub.ml/classes/models/llama.html#adapters.LlamaAdapterModel.add_classification_head or https://docs.adapterhub.ml/classes/models/bert.html#adapters.BertAdapterModel.add_dependency_parsing_head.

So is this issue introduced by the changes in this PR?

calpt (approved):

calpt:
@lenglaender I've fixed the docs build issues

PR #655, opened 2024-03-05
Fix embedding Training Bug

hSterz:
Fixes a bug that was prevent embeddings from training properly and changes the test to catch that in the future.


calpt (approved):
lgtm

PR #663, opened 2024-03-31
Add support for QLoRA/ QAdapter training via bitsandbytes

calpt:
This PR adds support for wrapping bitsandbytes' `Linear4bit` and `Linear8bitLt` quantization layers with our LoRA implementation, enabling training LoRA adapters on quantized models in QLoRA style.

Implementation is loosely similar to HF peft's approach, which can be found here: https://github.com/huggingface/peft/blob/v0.10.0/src/peft/tuners/lora/bnb.py.

## Demo

I've added a new notebook here: https://github.com/calpt/adapter-transformers/blob/dev/qlora/notebooks/QLoRA_Llama_Finetuning.ipynb.
The notebook showcases this feature by finetuning a 4bit-quantized Llama 2 7B on an instruction tuning dataset (similar to Guanaco in the QLoRA paper).
Tested that it runs without errors in the provided notebook, other setups are not extensively tested yet.

## Pre-trained checkpoints

Adapters trained with the notebook code can be found here:

Llama-2 7B: https://huggingface.co/AdapterHub/llama2-7b-qlora-openassistant
Llama-2 13B: https://huggingface.co/AdapterHub/llama2-13b-qlora-openassistant

## Current limitations

- [ ] Auto-casting in adapter layer not implemented
- [x] Only supports LoRA currently (fix in #678 required for support of others)
- [ ] No support for QLoRA in MergedLinear layer


src/adapters/methods/lora.py
```diff
@@ -429,9 +447,12 @@ def compose_single(self, adapter_setup: str, state: LoRAState, lvl: int = 0) ->
...
     def forward(self, input_states: torch.Tensor):
-        weight = torch.transpose(self.weight, -2, -1) if self.fan_in_fan_out else self.weight
-        # result shape: <batch_size> x <seq_len> x <head_dim>
-        layer_output = F.linear(input_states, weight, bias=self.bias)
+        if self.fan_in_fan_out:
+            weight = torch.transpose(self.weight, -2, -1) if self.fan_in_fan_out else self.weight
+            # result shape: <batch_size> x <seq_len> x <head_dim>
+            layer_output = F.linear(input_states, weight, bias=self.bias)
+        else:
+            layer_output = super().forward(input_states)
```
hSterz, line 455: Which forward method is called here since this does not inherit from nn.Linear anymore?

hSterz (approved):
Looks good just one small question about something that is unclear to me

src/adapters/methods/lora.py
calpt: the subclasses of this (`LoRALinearTorch`, `LoRALinear4bit`, `LoRALinear8bitLt`),  inherit from different types of linear layers

PR #673, opened 2024-04-11
Fix Training Error with AdapterDrop and Prefix Tuning

### LINKED ISSUE #669: T5 AdapterDrop Prefix-Tuning Bug
Hi, 

I'm trying to use train an AdapterDrop with T5. It's working fine for bottleneck, LoRA and IA3. But it does not work on  Prefix Tuning (Therefore MAM and UniPELT).

Following is a notebook: 
https://gist.github.com/FahadEbrahim/66686814f02978da9d4376470356647d 

The error is:
RuntimeError: The size of tensor a (90) must match the size of tensor b (80) at non-singleton dimension 3.

TimoImhof:
Fixes #669 

Changes in this PR:
- Avoid throwing `RuntimeError` due to dimension mismatch occuring when passing the positional encoding from layers dropped by AdapterDrop to layers modified by prefix tuning.

### DIFF
--- a/src/adapters/models/t5/modeling_t5.py
+++ b/src/adapters/models/t5/modeling_t5.py
@@ -37,6 +37,7 @@ from .mixin_t5 import (
     T5StackAdaptersMixin,
 )
 
+
 logger = logging.get_logger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


src/adapters/models/t5/modeling_t5.py
```diff
@@ -144,7 +144,10 @@ def project(hidden_states, proj_layer, key_value_states, past_key_value):
             query_states, key_states.transpose(3, 2)
         )  # equivalent of torch.einsum("bnqd,bnkd->bnqk", query_states, key_states), compatible with onnx op>9
 
-        if position_bias is None:
+        # For Prefix Tuning, when training with AdapterDrop, we must additionally check if the dimensions of the
+        # position_bias given from previous layers match the dimensions of the scores of the current layer to make
+        # sure that the position_bias is adequately recomputed if previous layers have been skipped.
+        if position_bias is None or position_bias.shape != scores.shape:
```
calpt, line 150: can we specifically check the seqlen dimensions here to make this condition more precise?

calpt (approved):
thanks!

PR #678, opened 2024-04-14
Fix compatibility of adapters with HF Accelerate auto device-mapping

calpt:
Adapters currently does not work correctly with passing `device_map="auto"` in a model's `from_pretrained()`. Device auto-mapping is handled by HF accelerate, which wraps the original module forward method.

This PR fixes compatibility of Adapters' post-hoc model wrapping with Accelerate's device auto-mapping via wrapping the forward pass.

Fixing this is required for enabling quantized training of adapters (bottleneck & prefix-tuning) in #663.

The fix can be tested like this:

```python
import adapters
from transformers import AutoTokenizer, AutoModelForSequenceClassification

model_name = "t5-small"

tokenizer = AutoTokenizer.from_pretrained(model_name)
model = AutoModelForSequenceClassification.from_pretrained(
    model_name,
    device_map="auto"
)

adapters.init(model)
model.add_adapter("my_adapter")
model.train_adapter("my_adapter")

model.to("cuda")

inputs = tokenizer("Hello, my dog is cute", return_tensors="pt")
inputs = inputs.to("cuda")

is_accessed = False

def hook_fn(module, args, output):
    global is_accessed
    is_accessed = True
    return output

adapter = model.get_adapter("my_adapter")
first_layer_module = adapter[0]["output_adapter"]
first_layer_module.register_forward_hook(hook_fn)

outputs = model(**inputs)
print(outputs)

assert is_accessed
```


hSterz (approved):
Looks good to me

Issue #680, opened 2024-04-15
Custom Heads not working with adapters


san-deep-reddy:
### Discussed in https://github.com/adapter-hub/adapters/discussions/679

<div type='discussions-op-text'>

<sup>Originally posted by **san-deep-reddy** April 14, 2024</sup>
I have tried many models and adapter types, custom head types and configs but I always end up with the same error -                                                    

```
 File "testing\adapter2_testing.py", line 173, in <module>
    trainer.train()
  File "C:\Users\dlais\AppData\Local\Programs\Python\Python39\lib\site-packages\transformers\trainer.py", line 1537, in train
    return inner_training_loop(
  File "C:\Users\dlais\AppData\Local\Programs\Python\Python39\lib\site-packages\transformers\trainer.py", line 1772, in _inner_training_loop
    self.control = self.callback_handler.on_train_begin(args, self.state, self.control)
  File "C:\Users\dlais\AppData\Local\Programs\Python\Python39\lib\site-packages\transformers\trainer_callback.py", line 370, in on_train_begin
    return self.call_event("on_train_begin", args, state, control)
  File "C:\Users\dlais\AppData\Local\Programs\Python\Python39\lib\site-packages\transformers\trainer_callback.py", line 414, in call_event
    result = getattr(callback, event)(
  File "C:\Users\dlais\AppData\Local\Programs\Python\Python39\lib\site-packages\transformers\integrations\integration_utils.py", line 635, in on_train_begin
    model_config_json = model.config.to_json_string()
  File "C:\Users\dlais\AppData\Local\Programs\Python\Python39\lib\site-packages\transformers\configuration_utils.py", line 951, in to_json_string
    return json.dumps(config_dict, indent=2, sort_keys=True) + "\n"
  File "C:\Users\dlais\AppData\Local\Programs\Python\Python39\lib\json\__init__.py", line 234, in dumps
    return cls(
  File "C:\Users\dlais\AppData\Local\Programs\Python\Python39\lib\json\encoder.py", line 201, in encode
    chunks = list(chunks)
  File "C:\Users\dlais\AppData\Local\Programs\Python\Python39\lib\json\encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "C:\Users\dlais\AppData\Local\Programs\Python\Python39\lib\json\encoder.py", line 405, in _iterencode_dict
    yield from chunks
[...]

PR #700, opened 2024-05-13
Move custom head dict out of config

hSterz:
To make the model_config serializable and prevent the error mentioned in #680 move the costum_heads dictionary out of the config and make it a separate attribute of the model class.


calpt (commented):
Looks good in general. Will this induce any breaking changes in the library?

hSterz:
The only thing that I can think of that this could affect is the loading of a model with a custom head. However, I think the custom_head class needs to be readded in the previous implementation as well. Am I missing something?

calpt:
Sounds good, was just curious if this had any side effects I'm not aware of. Thanks!

calpt (approved):

PR #698, opened 2024-05-08
Add support for Task Arithmetics

lenglaender:
This PR adds support for various task arithmetic options for LoRA. Until now, our library supported averaging only by linearly combining different adapters. This may be insufficient, especially for LoRA — hence, several publications have proposed other ways to perform task arithmetic.

This PR:
- makes it easier to implement different weighting methods
- adds 2 additional merging methods for LoRA following these papers
- adds method to merge heads
- Docu & notebook

PR #705, opened 2024-05-31
Add ReFT (LoReFT, NoReFT, DiReFT)

calpt:
This PR integrates multiple ReFT variants as new adapter methods.

Paper: https://arxiv.org/pdf/2404.03592
Original code: https://github.com/stanfordnlp/pyreft

## Changes

- Add ReFT module implementation via `ReftLayer`, integrated into all models supported by Adapters. Integration via `init_reft()` method & Pytorch hook.
- Add new `ReftConfig` as base config class with three default instances: `LoReftConfig`, `NoReftConfig` and `DiReftConfig`.
- Method documentation can be found here: https://github.com/adapter-hub/adapters/blob/6c19ea06c143621a735226e477bf772068e55be3/docs/methods.md#reft

## Compatibility

Tested that Pyreft & Adapters produce the same outputs on inference by converting Pyreft checkpoints to Adapters checkpoints (tested settings: LoReft, NoReft, DiReft, weight tying, prefix, suffix, rank, mostly using roberta-base).

Script for testing & checkpoint conversion here: https://github.com/calpt/pyreft/blob/main/compatibility.py.

## Evaluation

Roberta-base with LoReFT on GLUE, using hyperparameters similar to the paper:

Task | Score
--- | ---
Cola (Matthews Corr.) | 53.95
MNLI (Acc.) | 83.23
MRPC (F1) | 91.70
QNLI (Acc.) | 90.94
QQP (Acc.) | 86.82
RTE (Acc.) | 76.53
SST-2 (Acc.) | 93.81
STS-B (Spearmanr) | 88.99

## Todos

- [x] Modeling implementations
- [x] Add test methods
- [x] Make all checks passing
- [x] Add documentation
- [x] Make sure implementation produces same outputs as original code
- [x] Sanity check training runs


frankaging:
@calpt Thanks for the PR! I took a quick look, and it looks promising. Here are two minor questions:
1. It seems like tied and untied weights are handled here:
https://github.com/adapter-hub/adapters/pull/705/files#diff-e791dd9c62ff127d32170821d7571b69e641f7ecc9462a58227fa5a80f3502f1R80

Does this mean if I want united weights among prefix and suffix, I will create two adaptors and set `self.prefix_positions` or `self.suffix_positions` to none?

2. Since in the ReFT paper, we only run experiments where we add interventions to the residual stream (e.g., transformer layer or block output), will, by default, `ReftConfig` assume this? If I want to add interventions to different streams (e.g., attention output, MLP up-projection), how will the `ReftConfig` look like?

Thanks!

calpt:
@frankaging Thanks for looking over this! Re your questions:

Re 1: In the current implementation, one or two modules per layer will be created depending on whether the `tied_weights` attribute in the config is set or not. See here:
https://github.com/adapter-hub/adapters/blob/25f1d1c03ff41547829329258d568498b0482735/src/adapters/methods/reft.py#L50-L62
This makes it very easy for a user to to tie or not tie weights when adding a single Reft adapter, e.g.:
```python
from adapters import AutoAdapterModel, ReftConfig

model = AutoAdapterModel.from_pretrained("...")

config = ReftConfig(
    layers="all", prefix_positions=1, suffix_positions=1, r=1,
    tied_weights=True  # set to True or False to share weights
)
model.add_adapter("my_reft", config=config)
model.set_active_adapters("my_reft")
```

Re 2: Currently, the Reft implementation always assumes interventions are added to the residual stream as you explained, since this is the method proposed in the paper. This is done via a PyTorch hook here:
https://github.com/adapter-hub/adapters/blob/25f1d1c03ff41547829329258d568498b0482735/src/adapters/methods/reft.py#L162-L169
While no other intervention points are added for now, we can easily extend with similar hooks for other intervention points where it makes sense to do so.

Thanks again for looking over this! Please let us know if you have any suggestions or ideas what we should add or change for the first version!
Currently, the PR is still in a draft state. Once remaining issues are fixed and we have some documentation, would be happy to get your feedback again before we merge.

frankaging:
@calpt Thanks for your responses! It makes sense to me.

Will the hook work out of the box for accelerated training (e.g., deepspeed, etc..)? Any existing tests on this? Thanks!

calpt:
@frankaging No extensive tests for training yet at this point. Deepspeed support of this library is unfortunately flaky in general and not really a focus at the moment, but using e.g. torch distributed or HF Accelerate should work in the end.

src/adapters/methods/reft.py
```diff
@@ -23,7 +23,7 @@ def __init__(
         self.orthogonal = orthogonal
         self.learned_source = nn.Linear(in_dim, r_dim, bias=True)
 
-        projection = nn.Linear(in_dim, r_dim)
+        projection = nn.Linear(in_dim, r_dim, bias=False)
         if orthogonal:
             nn.init.orthogonal_(projection.weight)
```
frankaging, line 28: nit: we might not need this as registering line below might trigger this call internally: https://github.com/pytorch/pytorch/blob/main/torch/nn/utils/parametrize.py#L421

i am also removing this line in the `pyvene` library -- kudos to `peft` library PR: https://github.com/huggingface/peft/pull/1657#discussion_r1570154894

src/adapters/methods/adapter_layer_base.py
```diff
@@ -133,16 +146,38 @@ def enable_adapters(self, adapter_setup: AdapterCompositionBlock, unfreeze_adapt
...
+
+    def freeze_adapter(self, adapter_name: str, freeze: bool = True):
+        """Freezes/ unfreezes an adapter module.
+
+        Args:
+            adapter_name (str): The name of the adapter to freeze/ unfreeze.
+            freeze (bool, optional): Whether to freeze the adapter. Defaults to True.
+        """
+        if adapter_name in self.refts:
+            self.refts[adapter_name].train(not freeze)
```
hSterz, line 163: Why do some of the methods of the Base Class use `self. reft` while others use `self.adapter_modules`?

src/adapters/methods/reft.py
```diff
@@ -0,0 +1,184 @@
...
+    if isinstance(output, torch.Tensor):
+        return module.reft_layer(output)
+    else:
+        return (module.reft_layer(output[0]),) + output[1:]
+
+
+def init_reft(model):
+    for _, layer in model.iter_layers():
+        if not hasattr(layer, "reft_layer"):
+            layer.reft_layer = ReftLayer(model.config, model.adapters_config)
```
hSterz, line 183: Where is the `reft_layer` attribute added to the modules?

tests/methods/test_reft.py
```diff
@@ -0,0 +1,78 @@
+from adapters import DiReftConfig, LoReftConfig
+from transformers.testing_utils import require_torch
+
+from .base import AdapterMethodBaseTestMixin
+
+
+@require_torch
+class ReftTestMixin(AdapterMethodBaseTestMixin):
+    reft_configs_to_test = [
+        (LoReftConfig(), ["refts.{name}."]),
```
hSterz, line 10: Should we maybe include a config with a suffix as well? With the logic of finding the last token, I think this might be more prone to breaking than the prefix.

src/adapters/model_mixin.py
```diff
@@ -968,6 +972,17 @@ def forward_context(self, context: ForwardContext, *args, **kwargs):
         if hasattr(self.base_model, "prefix_tuning"):
             context.prefix_states = self.base_model.prefix_tuning(*args, **kwargs)
 
+        # TODO this does not support padding on the left
```
hSterz, line 975: Do we want to leave this TODO open?

hSterz (commented):
This looks great! Just some small comments and questions

src/adapters/methods/reft.py
calpt: It's added in this line and nowhere else. Is there something to change about this logic?

tests/methods/test_reft.py
calpt: yep, added

src/adapters/methods/reft.py
hSterz: Ah nevermind, I have got the logic now

src/adapters/model_mixin.py
calpt: added, pls check if this is correct

docs/methods.md
```diff
@@ -295,3 +295,50 @@ model.add_adapter("dummy", config=config)
 _Papers:_
 - [The Power of Scale for Parameter-Efficient Prompt Tuning](https://aclanthology.org/2021.emnlp-main.243/) (Lester et al., 2021)
 
+## ReFT
+
+_Configuration class_: [`ReftConfig`](adapters.ReftConfig)
+
+Representation Fine-Tuning (ReFT) as first proposed by [Wu et al. (2024)](https://arxiv.org/pdf/2404.03592) leverages so-called interventions to adapt the pre-trained representations of a language model.
```
lenglaender, line 302:
```suggestion
Representation Fine-Tuning (ReFT), as first proposed by [Wu et al. (2024)](https://arxiv.org/pdf/2404.03592), leverages so-called interventions to adapt the pre-trained representations of a language model.
```

docs/methods.md
```diff
@@ -295,3 +295,50 @@ model.add_adapter("dummy", config=config)
...
+config = ReftConfig(
+    layers="all", prefix_positions=3, suffix_positions=0, r=1, orthogonality=False, subtract_projection=False
+)  # equivalent to DireftConfig()
+```
+
+In addition, _Adapters_ supports configuring multiple hyperparameters tuned in the ReFT paper in `ReftConfig`, including:
+- number of prefix positions, via `prefix_positions`
+- number of suffix positions, via `suffix_positions`
+- which layers to intervene on, via `layers`. This can either be `"all"` or a list of layer ids
+- whether to tie parameters between prefixes and suffixes, via `tied_weights`
```
lenglaender, line 340: Just a bit cleaner look:
```suggestion
- `prefix_positions`: number of prefix positions
- `suffix_positions`: number of suffix positions
- `layers`: The layers to intervene on. This can either be `"all"` or a list of layer ids
- `tied_weights`: whether to tie parameters between prefixes and suffixes
```

src/adapters/methods/adapter_layer_base.py
```diff
@@ -103,28 +102,42 @@ def average_adapter(self, adapter_name: str, input_adapters: Dict[str, float]) -
...
-    @abstractmethod
     def delete_adapter(self, adapter_name: str):
         """Deletes an adapter module from the layer.
 
         Args:
             adapter_name (str): The name of the adapter to delete.
         """
-        raise NotImplementedError()
+        if adapter_name in self.adapter_modules:
+            del self.adapter_modules[adapter_name]
```
lenglaender, line 133: Now that we have a default implementation of `delete_adapter`, `enable_adapters` and `get_adapter` you can remove the functions from the other adapter classes (LoRA, Bottleneck, ...) when they use this default implementation.
You don't have to remove the duplicate code in the classes for the `average_adapter` function as I will do this in #698

tests/methods/test_reft.py
lenglaender, line 1: Do you still plan to support left padding (you reverted it in your last commit)? In this case, I believe we need a test for it.

lenglaender (commented):
Commented on some minor things; everything else looks good & correctly implemented to me

Once the open comments are resolved & left padding is implemented this is ready to merge.

src/adapters/methods/reft.py
```diff
@@ -0,0 +1,184 @@
...
+        suff_ids = last_non_padding - self.suffix_positions + 1
+        prefix = hidden_states[:, :pref_ids, :]
+        suffix = []
+        for i, suff_id in enumerate(suff_ids):
+            suffix.append(hidden_states[i, suff_id : suff_id + self.suffix_positions, :])
+        suffix = torch.stack(suffix, dim=0)
+        if self.tied_weights:
+            adapted_states = [torch.cat([prefix, suffix], dim=1)]
+        else:
+            adapted_states = [prefix, suffix]
```
lenglaender, line 85: Regarding optimizing the reft implementation:
We often have 0 suffix_positions (at least zero is the default in the configs). Couldn't we skip these calculations in these cases?

In general, I don't see how we could optimize `ReftUnit` anymore. So the only methods to optimize are `_gather_adapted_states` and `_scatter_adapted_states`.

src/adapters/methods/adapter_layer_base.py
calpt: good point, was planning to do that! now done

lenglaender (approved):
Looks good to me. This is ready to merge

calpt:
Thanks! I've added some quick training results on GLUE tasks in the description, based on that the implementation looks good.

@frankaging re distributed training: I've tested it to work with torch distributed & HF Accelerate via the Trainer class, e.g. for GLUE:
```
torchrun --standalone --nnodes=1 --nproc-per-node=2 examples/pytorch/text-classification/run_glue.py \
  --model_name_or_path roberta-large \
  --task_name $TASK_NAME \
  --do_train \
  --do_eval \
  --max_seq_length $SEQ \
  --pad_to_max_length False \
  --per_device_train_batch_size 32 \
  --learning_rate $LR \
  --warmup_ratio $WARMUP \
  --num_train_epochs $EPOCH \
  --output_dir output/$TASK_NAME \
  --overwrite_output_dir \
  --train_adapter \
  --adapter_config "loreft[prefix_positions=$POS]"
```

frankaging:
> Thanks! I've added some quick training results on GLUE tasks in the description, based on that the implementation looks good.
> 
> @frankaging re distributed training: I've tested it to work with torch distributed & HF Accelerate via the Trainer class, e.g. for GLUE:
> 
> ```
> torchrun --standalone --nnodes=1 --nproc-per-node=2 examples/pytorch/text-classification/run_glue.py \
>   --model_name_or_path roberta-large \
>   --task_name $TASK_NAME \
>   --do_train \
>   --do_eval \
>   --max_seq_length $SEQ \
>   --pad_to_max_length False \
>   --per_device_train_batch_size 32 \
>   --learning_rate $LR \
>   --warmup_ratio $WARMUP \
>   --num_train_epochs $EPOCH \
>   --output_dir output/$TASK_NAME \
>   --overwrite_output_dir \
>   --train_adapter \
>   --adapter_config "loreft[prefix_positions=$POS]"
> ```

@calpt Thanks! It's great to see this approach works for different kinds of parallel training!

I was re-looking into that orthogonal matrix initialization thing (i.e., I was referencing the PEFT repo ticket and asking whether we should remove a redundant init), and I found that for some cases, removing that init step might cause unstable results. Have you look in to this again by doing some tests on your side? Thanks.

calpt:
> I was re-looking into that orthogonal matrix initialization thing (i.e., I was referencing the PEFT repo ticket and asking whether we should remove a redundant init), and I found that for some cases, removing that init step might cause unstable results. Have you look in to this again by doing some tests on your side? Thanks.

Interesting, I haven't tested this specifically. From looking at the code, it makes sense the orthogonal init is redundant, would you suggest we re-add it still?

PR #715, opened 2024-07-05
[DOCS] Fixed typos inside notebook files and doc pages

julian-fong:
@calpt I added some changes to notebook 3 as requested, and made some various typo fixes inside other notebooks. A review for notebook 3 would be appreciated (i edited something on cell 6)! This is also my first pr on `adapters` :)

Also it looks like the notebook metadata changed as well even when I did a very minor edit to some text

### DIFF
--- a/notebooks/Text_Generation_Training.ipynb
+++ b/notebooks/Text_Generation_Training.ipynb
@@ -17,20 +17,20 @@
       "cell_type": "code",
       "execution_count": 1,
       "metadata": {
+        "ExecuteTime": {
+          "end_time": "2023-08-17T12:15:51.447758300Z",
+          "start_time": "2023-08-17T12:15:46.491925400Z"
+        },
         "colab": {
           "base_uri": "https://localhost:8080/"
         },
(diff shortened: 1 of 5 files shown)


notebooks/Text_Generation_Training.ipynb
```diff
@@ -17,20 +17,20 @@
       "cell_type": "code",
```
calpt, line 17: I'm unsure if anything besides metadata has changed in this notebook that should be reviewed? If not, we could also revert changes here.

calpt (commented):
Thanks for working on this, changes look good to me overall!

notebooks/Text_Generation_Training.ipynb
julian-fong: In `Text_Generation_training` there was a cell that had an extra indent - so I just un-indented it just as a quick fix, but I realized that i didn't commit those changes in yet, will do so now

calpt (approved):

PR #716, opened 2024-07-08
Update training.md

Kotstantinovskiy:
Fix to TrainingArguments
TrainingsArguments doesn't exist in the transformers library


calpt (approved):
thanks for fixing!

PR #730, opened 2024-08-07
Don't throw error if ForwardContext is not available

calpt:
Fixes in this PR:
- Fix 'NoneType' error if context not available
- Fix Llama grad checkpointing training.
- Add quantized training test.

Still existing limitation:
- ForwardContext does not work with gradient checkpointing, ie methods such as reft or prefix tuning don't work with gradient checkpointing currently.


lenglaender (approved):
Looks good to me! One last thing: The newly added test is currently skipped because in our pipeline we don't yet install bitsandbytes:

image

As soon as you have added this to the pipeline & the tests run through, you can merge it

PR #736, opened 2024-08-20
Fix search on docs.adapterhub.ml

TheoWeih:
Updated `sphinx-rtd-theme==2.0.0`.

The styling after building it with `make clean && make html` looked the same to me,
but maybe a reviewer could confirm this.

Updating this package fixes the search bar.

image


calpt (approved):
thanks for fixing!

PR #738, opened 2024-08-25
Pluggable Model Integration Interface

calpt:
This PR drafts a new model integration interface which makes it easier to support new and custom model architectures for selected adapter methods without full model implementation.

This is done with the new `AdapterModelInterface` class that translates from generic model access points to model-specific attribute names.

### Example usage:

Basic interface for Qwen model:
```python
model_interface = AdapterModelInterface(
    adapter_types=["lora", "reft"],
    model_embeddings="embed_tokens",
    model_layers="layers",
    layer_self_attn="self_attn",
    layer_cross_attn=None,
    attn_k_proj="k_proj",
    attn_q_proj="q_proj",
    attn_v_proj="v_proj",
    attn_o_proj="o_proj",
    layer_intermediate_proj="mlp.up_proj",
    layer_output_proj="mlp.down_proj",
)

model_name = "Qwen/Qwen2-0.5B"

model = AutoModelForCausalLM.from_pretrained(model_name)
adapters.init(model, interface=model_interface)

config = LoRAConfig()
# config = LoReftConfig()
model.add_adapter("my_adapter", config=config)

print(model.adapter_summary())
```

#### Extended interface

Additionally, the interface provides optional attributes that enable (almost) full bottleneck adapter support. Without the extended interface, bottleneck adapter support is very limited.

Example for Gemma2:

```python
adapter_interface = AdapterModelInterface(
    adapter_types=["bottleneck", "lora", "reft"],
    model_embeddings="embed_tokens",
    model_layers="layers",
    layer_self_attn="self_attn",
    layer_cross_attn=None,
    attn_k_proj="k_proj",
    attn_q_proj="q_proj",
    attn_v_proj="v_proj",
    attn_o_proj="o_proj",
    layer_intermediate_proj="mlp.up_proj",
    layer_output_proj="mlp.down_proj",
    layer_pre_self_attn="input_layernorm",
    layer_pre_cross_attn=None,
    layer_pre_ffn="pre_feedforward_layernorm",
    layer_ln_1="post_attention_layernorm",
    layer_ln_2="post_feedforward_layernorm",
)
```

### Additional novelties

- Adds `AdapterMethod` as an enum of all supported adapter method types (e.g. `AdapterMethod.bottleneck`, `AdapterMethod.lora`, ...)
- Adds a `supports_adapter()` method for easy checking whether a model instance supports a certain adapter method. This method can receive an `AdapterMethod` string or a config object:
    ```python
    model.supports_adapter(AdapterMethod.prompt_tuning)
    # or
    model.supports_adapter(PromptTuningConfig())
    ```
    (This method is supported by both models implemented via "classic" mixins and via pluggable interface.)

### State of implementation

Supported adapter types:
- [x] LoRA
- [x] ReFT
- [x] Bottleneck/ Compacter: **partial**, currently does **not** support:
    - [x] `is_parallel`, via extended interface
    - [x] `original_ln_before=True`, via extended interface
    - [ ] `original_ln_after=False` (e.g. used for `AdapterPlusConfig`)
- [x] Invertible adapters
- [ ] Prefix Tuning
- [x] Prompt Tuning: **partial**: attention mask modification only supports very specific model implementations

Supported features:
- [x] Embedding training
- [x] Fusion composition

**Not** to be supported:
- Parallel composition
- AdapterModel classes


docs/model_overview.md
```diff
@@ -13,6 +13,7 @@ The table below further shows which model architectures support which adaptation
 
 | Model                                   | (Bottleneck)<br> Adapters | Prefix<br> Tuning | LoRA | Compacter | Adapter<br> Fusion | Invertible<br> Adapters | Parallel<br> block | Prompt<br> Tuning | ReFT |
 | --------------------------------------- | -| - | - | - | - | - | - |- | - |
+| [Custom models](plugin_interface.html)    | ✅(°) |  | ✅ | ✅ | ✅ | ✅ |  | ✅ | ✅ |
```
lenglaender, line 16: Once we update this PR to include some model interfaces out of the box (like Gemma), we should list them here too and mention them in docs/plugin_interface.md

I believe this is a blocker for merging - we need to integrate these model interfaces before we merge.

src/adapters/interface.py
lenglaender, line 1: I don't like that one must always specify all attributes _(except the ones used for bottleneck)_ even if we support only a subset of the adapter methods. For example, if I only want to use ReFT and thus set `adapter_methods=["reft"]`, I still have to set the params needed for LoRA (`attn_k_proj`, ...) or else I get an error.

lenglaender (commented):
Looks really promising; really nice that we have this feature now! 🎉

docs/model_overview.md
calpt: I think it's perfectly fine to provide the built-in interfaces in a separate PR. This way we also don't make a single huge change as this PR is already quite big.

src/adapters/interface.py
lenglaender: Makes sense to keep it like it currently is to avoid extra complexity. So let's stick with the current approach :+1:

calpt:
Remaining test failure will be fixed by #789.

lenglaender (approved):
Looks good to me! Once the remaining tests are fixed, this can be merged

PR #740, opened 2024-09-16
Tests Refactoring

### LINKED ISSUE #785: ConfigUnion test failures: Identical outputs with and without adapters when combining CompacterConfig and LoRAConfig
## Initial Problem: Incomplete Test Coverage
During test refactoring, I discovered that the ConfigUnion test class only tests the first of three configs. This occurs because the model is created once and reused across subtests without removing previously added adapters.

Initially, I fixed this by removing adapters at the end of each test to ensure a clean model state for subsequent subtests.
However, this revealed a deeper issue in the `test_union_adapter_forward` test:

```
def run_forward_test(self, model, adapter_config, dtype=torch.float32):
    model.eval()

    name = adapter_config.__class__.__name__
    if name not in model.adapters_config:
        model.add_adapter(name, config=adapter_config)
    model.to(torch_device).to(dtype)

    input_data = self.get_input_samples(config=model.config, dtype=dtype)

    # pass 1: set adapter via property
    model.set_active_adapters(name)
    output_1 = model(**input_data)

    # pass 2: set via context
    # unset and make sure it's unset
    model.set_active_adapters(None)
    self.assertEqual(None, model.active_adapters)
    with AdapterSetup(name):
        output_2 = model(**input_data)

    # pass 3: base output
    model.set_active_adapters(None)
    base_output = model(**input_data)

    self.assertEqual(len(output_1), len(output_2))
    self.assertTrue(torch.equal(output_1[0], output_2[0]))
    self.assertGreaterEqual(len(output_1), len(base_output))
    self.assertFalse(torch.equal(output_1[0], base_output[0]))

    # Initial Fix: Remove added adapters in case of multiple subtests
    model.set_active_adapters(None)
    model.delete_adapter(name)
```
## Unexpected Behavior After Fixing
In this test, we compare the model forward passes when active adapters are used vs. when not. 
After applying the initial fix on this test (so we now also test the residual two configs), this revealed the following issue for some models for the config below:

```
# config:
(
    ConfigUnion(
        CompacterConfig(phm_dim=1),
        LoRAConfig(),
    ),
    ["adapters.{name}.", "loras.{name}."],
),

# error:
self.assertFalse(torch.equal(output_1[0], base_output[0]))
E   AssertionError: True is not false

```
This means that there is no difference between the `base_output` and `output_1`.
After spending some time on debugging this, I was not able to find the root cause for this problem, hence I am creating this dedicated issue. For now I reverted the fix in the `run_forward_test` and will link this issue in the code base to provide context.

TimoImhof:
This PR aims to refactor the test suite into a more generic version.

The following points are to be covered (subject to change; will become more granular):
- [x] Add more granularity to test viewer for better test overview/readability and more convenient testing of subgroups of methods (e.g. only exectue LoRA tests for one model)
- [x] Add documentation about the structure of the test directory and entry points
- [x] Add pytest markers to allow for more granular testing of subgroups (e.g. exectue only LoRA tests for all models at once, useful when implementing a new adapter method)
- [x] Refactor test methods to extract similar/duplicate code into a set of utils methods
 - [x] Relocate edgecases due to model peculiarities out of the tests and into the respective model test classes to keeps the tests generic
 - [x] Fix config union tests, closes #785 

This should make it easier to add new models or methods to the existing test suite in the future and make the development/testing process more convenient in the future


tests/test_methods/method_test_impl/base.py
```diff
@@ -56,11 +48,15 @@ def run_add_test(self, model, adapter_config, filter_keys):
         # check that weights are available and active
         has_weights = False
         filter_keys = [k.format(name=name) for k in filter_keys]
-        for k, v in self.filter_parameters(model, filter_keys).items():
+        for k, v in self._filter_parameters(model, filter_keys).items():
             has_weights = True
             self.assertTrue(v.requires_grad, k)
         self.assertTrue(has_weights)
 
+        # TODO: for config union tests resetting model should lead to that adapter can be readded with the same name, but currently not working
```
calpt, line 56: this implies there being a bug with configunion?

tests/test_methods/method_test_impl/base.py
```diff
@@ -160,6 +156,7 @@ def run_forward_test(self, model, adapter_config, dtype=torch.float32):
         model.eval()
 
         name = adapter_config.__class__.__name__
+        # TODO: this defeats the purpose of the test, for the config union tests as only the first config is added
```
calpt, line 159: so we should rewrite the test for configunion?

tests/test_methods/generator.py
```diff
@@ -0,0 +1,218 @@
...
+    class Core(
+        model_test_base,
+        CompabilityTestMixin,
+        AdapterFusionModelTestMixin,
+        unittest.TestCase,
+    ):
+        pass
+
+    if "Core" not in excluded_tests:
+        test_classes["Core"] = Core
```
calpt, line 60: minor: since we only ever use the class if the if-clause is true, we could define it in the if-clause?

setup.cfg
```diff
@@ -49,6 +49,9 @@ use_parentheses = True
 [flake8]
 ignore = E203, E501, E731, E741, W503, W605
 max-line-length = 119
+per-file-ignores =
+    tests/test_methods/generator.py: F401, F403, F405
+    tests/test_methods/test_*.py:F403,F405
```
calpt, line 54: why are these exceptions needed?

calpt (commented):
Looks really great overall! Also really appreciate the detailed readme!

Left a couple of comments mainly for clarification

tests/test_methods/method_test_impl/core/test_adapter_conversion.py
```diff
@@ -110,23 +110,7 @@ def test_conversion_seq2seq_lm_model(self):
...
-                0, model.config.vocab_size, size=self.default_input_samples_shape[:-1], device=torch_device
-            )
-            label_dict["labels"] = label_dict["decoder_input_ids"]
-        else:
-            model = MODEL_FOR_SEQ_TO_SEQ_CAUSAL_LM_MAPPING[self.config_class](self.config())
-            label_dict["labels"] = torch.zeros(
-                (self.batch_size, self.seq_length), dtype=torch.long, device=torch_device
-            )
-            label_dict["decoder_input_ids"] = label_dict["labels"].clone()
+        model, label_dict = self.get_conversion_model()
```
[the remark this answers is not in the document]
TimoImhof: That's a good observation!
I agree with you, I created the respective methods and removed the get_conversion_model() method

tests/test_methods/test_on_albert.py
```diff
@@ -0,0 +1,48 @@
...
+
+method_tests = generate_method_tests(AlbertAdapterTestBase)
+
+for test_class_name, test_class in method_tests.items():
+    globals()[test_class_name] = test_class
+
+
+@require_torch
+@pytest.mark.heads
+class Heads(
```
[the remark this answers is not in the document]
TimoImhof: Until now it is, I did not find a good workaround to overwrite methods that are specific to only some of the generated test classes, especially when super().method() is called in the overwritten method.

And yes good catch, we can exclude it beforehand to save runtime 👍

tests/test_methods/test_on_clip/test_model.py
```diff
@@ -0,0 +1,79 @@
...
+        # This test is not applicable to CLIP
+        self.skipTest("Not applicable to CLIP.")
+
+    def test_load_adapter_setup(self):
+        self.skipTest("Not applicable to CLIP.")
+
+
+method_tests = generate_method_tests(
+    model_test_base=CLIPAdapterTestBase,
+    excluded_tests=["Embeddings", "Heads", "Composition", "ClassConversion", "PromptTuning", "ConfigUnion"],
```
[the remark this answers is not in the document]
TimoImhof: Yes all Clip test bases inherited only from these mixins:

``` 
    BottleneckAdapterTestMixin,
    CompacterTestMixin,
    IA3TestMixin,
    LoRATestMixin,
    PrefixTuningTestMixin,
    ReftTestMixin,
    UniPELTTestMixin,
    AdapterFusionModelTestMixin,
    CompabilityTestMixin,
```

Where Compatibility and Fusion Mixins are summarized in the new Core category.

tests/test_methods/test_on_deberta.py
```diff
@@ -0,0 +1,27 @@
...
+        num_attention_heads=4,
+        intermediate_size=37,
+        hidden_act="gelu",
+        relative_attention=True,
+        pos_att_type="p2c|c2p",
+    )
+    tokenizer_name = "microsoft/deberta-base"
+
+    def test_parallel_training_lora(self):
+        self.skipTest("Not supported for DeBERTa")
```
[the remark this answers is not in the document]
TimoImhof: I fixed the GPT2 file. 
The mechanism in Deberta is the intended way

setup.cfg
TimoImhof: These are because all test imports are bundled in the generator.py file and our `make quality` command would otherwise detect unused or from star imports there

tests/test_methods/generator.py
TimoImhof: I think that's a valid point and also more efficient 👍

tests/test_methods/test_on_clip/test_model.py
```diff
@@ -0,0 +1,79 @@
+import random
```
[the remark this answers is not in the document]
TimoImhof: Generating a test class multiple times in one file would lead to overwriting the same test class over and over, so right now this is not possible.

Alternatively as you suggested one could omit the parent folder and just add separate files for each type of CLIP model.
However one of the main goals was improved structure in the test viewer, which currently displays the CLIP tests like this because we use a parent folder for CLIP:

image

(Which I find very practical)

When removing it we would loose this granularity, which in my opinion brings more practical value when working with the tests and outweighs the minor inconsistency in file organization.

tests/test_methods/test_on_mbart.py
```diff
@@ -0,0 +1,29 @@
...
+        vocab_size=250027,
+    )
+    tokenizer_name = "facebook/mbart-large-cc25"
+
+    def test_parallel_training_lora(self):
+        self.skipTest("Not supported for MBart")
+
+
+method_tests = generate_method_tests(
+    MBartAdapterTestBase, excluded_tests=["ConfigUnion", "Embeddings", "PromptTuning"]
```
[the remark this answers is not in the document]
TimoImhof: They are excluded because they were not included previously:

See here:
```
@require_torch
class MBartAdapterTest(
    BottleneckAdapterTestMixin,
    CompacterTestMixin,
    IA3TestMixin,
    LoRATestMixin,
    PrefixTuningTestMixin,
    ReftTestMixin,
    UniPELTTestMixin,
    AdapterFusionModelTestMixin,
    PredictionHeadModelTestMixin,
    ParallelAdapterInferenceTestMixin,
    MBartAdapterTestBase,
    unittest.TestCase,
):
    pass
```

If they were originally not included due to redundancy I agree with you that we could use an extra comment to acknowledge this 👍

tests/test_methods/test_on_clip/test_model.py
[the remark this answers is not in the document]
calpt: alright, no strong opinion on this, so we can keep as is 👍

setup.cfg
calpt: I see, but why is it necessary to use this kind of import structure? flake8 and our quality checks try to enforce these import styles whenever possible, but maybe I'm missing why the usual rules would be impractical here?

TimoImhof: It is not strictly necessary, I mainly thought of this as a way to dodge manually importing all necessary modules in all testfiles.
But this is a small fix and we can revert to normal imports if we want to strictly enforce the import styles, I don't have a strong opinion here

tests/test_methods/method_test_impl/peft/test_config_union.py
```diff
@@ -16,31 +16,32 @@ class ConfigUnionAdapterTest(AdapterMethodBaseTestMixin):
...
             ["adapters.{name}.", "loras.{name}."],
         ),
     ]
 
     def test_add_union_adapter(self):
-        model = self.get_model()
-        model.eval()
-
+        # TODO: Discuss, why old tests were not working properly (could not work because we would add three times the same adapter name)
+        # TODO: Discuss why these config unions are not working properly (must set phm_dim=1)
```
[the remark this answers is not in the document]
TimoImhof: Yes I discovered that our config union tests are not working properly because only the first of the three configs is tested. This happens because we create the adapter name from the class of the config, which in this case always is "ConfigUnion".

To circumvent this now every subTest has its own model instance, which revealed that the current test fail for the remaining two configs:

```
            ConfigUnion(
                SeqBnConfig(),
                LoRAConfig(),
            ),
            ["adapters.{name}.", "loras.{name}."],
        ),
        (
            ConfigUnion(
                LoRAConfig(),
                CompacterConfig(),
            ),
            ["adapters.{name}.", "loras.{name}."],
        ),
```

the `test_add_union_adapter` failed due to this error:
```
AssertionError: Argument `out_features`=1 is not divisble be `phm_dim`4 or (out_features=2)
```
This could be fixed by just setting phm_dim to 1, which then revealed that this config:

```
        (
            ConfigUnion(
                LoRAConfig(),
                CompacterConfig(),
            ),
            ["adapters.{name}.", "loras.{name}."],
        ),
```
for some models does not produce a different output when no adapters are activated during the `test_union_adapter_forward` test.

I tried debugging this, however until now I did not find the cause for this behavior.

tests/test_methods/method_test_impl/base.py
TimoImhof: Yes that's right. My first idea was that if we just delete the existing adapter in the test model we should have no trouble adding a new one with the same name. However this did not work, which led me to the current solution of creating a new model in each subtest.

tests/test_methods/method_test_impl/base.py
TimoImhof: I don't think this is necessary, the current workaround with a new model for every subtest fixes this issue and when it becomes clear why the deleting of the adapter and then readding a new one with the same name does not work properly we can go back to reusing the same model and just deleting the adapter at the end of the `run_forward_test` method.

setup.cfg
TimoImhof: I removed the exceptions in the setup.cfg file and reverted the import structure to standard imports 👍

tests/test_methods/method_test_impl/peft/test_config_union.py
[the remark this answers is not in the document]
TimoImhof: I created a dedicated issue for this problem, see #785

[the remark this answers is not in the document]
TimoImhof: I think I solved it, the tests now work for all configs.
Details are provided in the issue #785.

PR #746, opened 2024-10-19
Add support for AdapterPlus

julian-fong:
This PR aims to add support for AdapterPlus

Github: https://github.com/visinf/adapter_plus
Paper: https://arxiv.org/pdf/2406.06820

Integration of AdapterPlus into the `adapters` library will involve adding new parameters/options to the `BnConfig`

Checklist of things that are added/to be added

1. New type of `scaling` called `channel`, in which we add learnable parameters for the channel/input_size dimension
2. New type of `init_weights` called `houlsby`, where the projection matrices $W_{down}$ and $W_{up}$ will be initialized with zero-centered Gaussian with a standard deviation of $10^{-2}$ truncated at 2 standard deviations, and zeros for bias
3. Support for `drop_path`, also known as stochastic depth **ONLY** applicable for vision based tasks using residual networks - located under a new file called `/methods/vision.py`


src/adapters/configuration/adapter_config.py
```diff
@@ -364,6 +370,24 @@ class ParBnConfig(BnConfig):
...
+
+    original_ln_after: bool = False
+    residual_before_ln: bool = True
+    stochastic_depth: float = 0.1
+    init_weights: str = "houlsby"
+    scaling: Union[float, str] = "channel"
+
+    mh_adapter: bool = False
+    output_adapter: bool = True
+    reduction_factor: Union[float, Mapping] = 2
```
calpt, line 387: does this value correspond to the config used for experiments in the paper? I believe they use rank 8. Our reduction factor value is relative to the hidden dim of the model (768 for ViT base), ie rank = hidden_dim / reduction_factor -> reduction_factor = 768 / 8

src/adapters/methods/vision.py
```diff
@@ -0,0 +1,66 @@
+# Module to support adapter training for vision related tasks
```
calpt, line 1: I believe it's fully fine to just use the implementation of torchvision here if there's no functional difference. We most likely require that package anyway for training vision models (e.g. for image transforms, see [here](https://github.com/huggingface/transformers/blob/1d063793318b20654ebb850f48f43e0a247ab7bb/examples/pytorch/image-classification/run_image_classification.py#L27)).

calpt (commented):
Did a first review pass, looks pretty solid overall! thanks for working on this!

Could you additionally add the new config to relevant locations in the docs. e.g. `methods.md` and `classes/adapter_config`?

src/adapters/configuration/adapter_config.py
julian-fong: Yes, you are correct. At the time I wasn't sure what `reduction_factor` to use since the paper specifies they wanted a rank of 8 instead of a specified `reduction_factor`. Is it correct to assume that we should just use `reduction_factor` of 96 i.e 768/96 = 8 since the user will be using the `ViT` model with a `hidden_dim` of 768, would that be the best practice in this scenario?

src/adapters/methods/vision.py
julian-fong: Sounds good! then i will import the `stochastic_depth` function from torchvision inside `src/adapters/methods/modeling.py`

julian-fong:
@calpt It looks like there is a missing `torchvision` import error happening in the CI. Is the CI configured to install the package? I don't see it inside as a core dependency in `setup.py` but not exactly sure if that is the correct place to look.

calpt:
> @calpt It looks like there is a missing `torchvision` import error happening in the CI. Is the CI configured to install the package? I don't see it inside as a core dependency in `setup.py` but not exactly sure if that is the correct place to look.

you're right, it's not currently installed in CI. we could add it as optional dependency in setup.py, similar to what HF does here: https://github.com/huggingface/transformers/blob/405b56269812056d9593869e22b7b264d806cb1e/setup.py#L312 and then install that additional dep package here https://github.com/adapter-hub/adapters/blob/bcace97bd3fb0db821780d4c2bcd087fbf200386/.github/workflows/tests_torch.yml#L66

julian-fong:
> > @calpt It looks like there is a missing `torchvision` import error happening in the CI. Is the CI configured to install the package? I don't see it inside as a core dependency in `setup.py` but not exactly sure if that is the correct place to look.
> 
> you're right, it's not currently installed in CI. we could add it as optional dependency in setup.py, similar to what HF does here: https://github.com/huggingface/transformers/blob/405b56269812056d9593869e22b7b264d806cb1e/setup.py#L312 and then install that additional dep package here
> 
> https://github.com/adapter-hub/adapters/blob/bcace97bd3fb0db821780d4c2bcd087fbf200386/.github/workflows/tests_torch.yml#L66

Would you like that in a separate pr? I know other open-source libraries prefer another pr to differentiate scopes of pull requests

julian-fong:
> > @calpt It looks like there is a missing `torchvision` import error happening in the CI. Is the CI configured to install the package? I don't see it inside as a core dependency in `setup.py` but not exactly sure if that is the correct place to look.
> 
> you're right, it's not currently installed in CI. we could add it as optional dependency in setup.py, similar to what HF does here: https://github.com/huggingface/transformers/blob/405b56269812056d9593869e22b7b264d806cb1e/setup.py#L312 and then install that additional dep package here
> 
> https://github.com/adapter-hub/adapters/blob/bcace97bd3fb0db821780d4c2bcd087fbf200386/.github/workflows/tests_torch.yml#L66

Please correct me if I'm wrong, but wouldn't `torchvision` need to be included as a core dependency since (and maybe there is an alternative that avoids this) we're importing the `torchvision` `stochactic_depth` function in `methods/modeling.py`? Users will be missing the library for that import whenever the file is being used if they install `adapters` straight up. 

Thats sort of why I originally wrote the native implementation in the library so that we don't need to include any new core dependencies.

calpt:
> > > @calpt It looks like there is a missing `torchvision` import error happening in the CI. Is the CI configured to install the package? I don't see it inside as a core dependency in `setup.py` but not exactly sure if that is the correct place to look.
> > 
> > 
> > you're right, it's not currently installed in CI. we could add it as optional dependency in setup.py, similar to what HF does here: https://github.com/huggingface/transformers/blob/405b56269812056d9593869e22b7b264d806cb1e/setup.py#L312 and then install that additional dep package here
> > https://github.com/adapter-hub/adapters/blob/bcace97bd3fb0db821780d4c2bcd087fbf200386/.github/workflows/tests_torch.yml#L66
> 
> Would you like that in a separate pr? I know other open-source libraries prefer another pr to differentiate scopes of pull requests, so just wondering what your preference is

fine to include here since this is directly related to the new functionality implemented in this pr

calpt:
> > > @calpt It looks like there is a missing `torchvision` import error happening in the CI. Is the CI configured to install the package? I don't see it inside as a core dependency in `setup.py` but not exactly sure if that is the correct place to look.
> > 
> > 
> > you're right, it's not currently installed in CI. we could add it as optional dependency in setup.py, similar to what HF does here: https://github.com/huggingface/transformers/blob/405b56269812056d9593869e22b7b264d806cb1e/setup.py#L312 and then install that additional dep package here
> > https://github.com/adapter-hub/adapters/blob/bcace97bd3fb0db821780d4c2bcd087fbf200386/.github/workflows/tests_torch.yml#L66
> 
> Please correct me if I'm wrong, but wouldn't `torchvision` need to be included as a core dependency since (and maybe there is an alternative that avoids this) we're importing the `torchvision` `stochactic_depth` function in `methods/modeling.py`? Users will be missing the library for that import whenever the file is being used if they install `adapters` straight up.
> 
> Thats sort of why I originally wrote the native implementation in the library so that we don't need to include any new core dependencies for `timm` or `torchvision`.

that's a good point. I think what we can do is to make the import of torchvision conditional on the package being available. this is pretty standard practice across Transformers and they have helper methods for checking all sorts of packages, including torchvision: https://github.com/huggingface/transformers/blob/33868a057c02f0368ba63bd1edb746be38fe3d90/src/transformers/utils/import_utils.py#L361

If torchvision is not installed & stochastic_depth is used, we should thrown an exception, otherwise we don't care whether it's installed. This keeps it optional for all non-vision use cases.

calpt (approved):
looks good to me overall, thanks a lot for working on this!
can be merged from my side once all checks are passing

julian-fong:
> looks good to me overall, thanks a lot for working on this! can be merged from my side once all checks are passing

I have fixed the code quality issues - thanks!

PR #750, opened 2024-10-23
Fix import error for Huggingface-hub version >=0.26.0 & Update Notebooks

lenglaender:
This PR solves an import error, which causes Adapters to not work with huggingface_hub version >= 0.26. This import error comes from HuggingFace removing deprecated functions.

Additionally, I updated all our example notebooks because they were outdated and used the `use_auth_token` parameter that isn't used by Hugging Face anymore.

## Context for Why This Error Occurred
To download files from repositories, we have the `download_cached(...)` and `get_from_cache(...)` functions in src/adapters/utils.py. Hugging Face has used similar methods in the past, too - our get_from_cache is copied from them. Now, Hugging Face uses a new method: `hf_hub_download(...)` (huggingface_hub/file_download.py)

However, we can't use the new `hf_hub_download(...)` function because it has the validate_hf_hub_args decorator that checks if the arguments are valid. This includes checking if the URL is a Hugging Face repo. Our function, however, also needs to download files from GitHub since we have the mapping between the old AdapterHub adapter paths to the new Hugging Face Hub paths in files that are in GitHub. Therefore, we cannot use the new method and have these methods still copied.

### DIFF
--- a/src/adapters/utils.py
+++ b/src/adapters/utils.py
@@ -382,6 +382,11 @@ def get_from_cache(
 
 
 def download_cached(url, checksum=None, checksum_algo="sha1", cache_dir=None, force_extract=False, **kwargs):
+    """
+    This method downloads a file and caches it.
+
+    For more information on why this is needed, refer to the explanation in this Pull Request: https://github.com/adapter-hub/adapters/pull/750
+    """
     if isinstance(url, Path):
         url = str(url)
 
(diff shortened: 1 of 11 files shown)


calpt (approved):
thanks for fixing this!

src/adapters/utils.py
```diff
@@ -180,6 +181,38 @@ def remote_file_exists(url):
     return r.status_code == 200
 
 
+# Copied from here: https://github.com/huggingface/huggingface_hub/blob/v0.25.0/src/huggingface_hub/file_download.py#L266
+def url_to_filename(url: str, etag: Optional[str] = None) -> str:
```
14H034160212, line 185: should not add this function back again. Use hf_hub_download() instead.
https://github.com/huggingface/huggingface_hub/releases

PR #754, opened 2024-10-27
Fix output_embeddings get/ set for empty active head

### LINKED ISSUE #742: something wrong with delete_head implementation
when you call delete_head() to remove a classification head which is the only head in your model, you can not call add_classification_head() to add a new head, because there will be a error showing the following error:

`
97     for head_name in self._active_heads:
98         if head_name in self.heads:
99             head = self.heads[head_name]
TypeError: 'NoneType' object is not iterable
`

I assume that you should set self._active_heads = [] if there is no other heads left.

calpt:
Fixes #742.


TimoImhof (approved):
LGTM

PR #758, opened 2024-11-05
fix flash attention for mistral. 

divyanshuaggarwal:
## fix flash attention for mistral. 

This pull request fixes flash attention forward method for mistral.


calpt (approved):
thanks for fixing!

PR #607, opened 2023-11-23
Adding Prompt Tuning to Missing Tables

lenglaender:
Adding Prompt Tuning to all tables where it was missing and fixing broken links in the README.

PR #763, opened 2024-12-01
Implements `Vera`

julian-fong:
This PR aims to implement Vera, which introduces trainable parameters `d` and `b` while keeps LoRA matrices `A` and `B` frozen, random, and shared across layers. 

I've opted to put the Vera implementation under the Lora implementation (like IA3)

Paper: https://arxiv.org/pdf/2310.11454

This PR includes:

- A new `Vera` implementation under the lora methods file
- A new adapter config named `VeraConfig`
- A shared parameter initialization function that takes in the `init_weights` argument set from the `VeraConfig` 
- A new criteria inside the `add_adapter` function inside the `LoRALayer` to check if we should use the `Vera` class. Currently, the criteria will check to see if the parameters from the `VeraConfig` `d` and `b` are of type `float`, its the most simple criteria I could think of at the moment. 
- A new parameter included added inside `add_adapter` inside the `LoRALayer` as suggested. It will now pass the name of the adapter inside the `__init__` function. 

Things to note:

In the original Vera paper, the decomposition matrices B and A are frozen and shared across layers. As suggested, I've opted to create these matrices similar to the `PHMLayer` implementation, using a new method named `init_shared_vera_parameters`. This function will take in the `init_weights` argument set from the `VeraConfig` to setup the initialization of B and A, while the other parameters from the `VeraConfig` will be used inside the `Vera` module. 

As I'm not 100% sure what the composition modes do ('add', 'scale'), I've opted the Vera class to be use only if `composition_mode` is set to `add`, and either `d` or `b` are of type float (and not `None`). 

reviews appreciated!


julian-fong:
@calpt Regarding the paper: from my first readings it seems like it doesn't mention any scaling via a constant `alpha/r` or any gating. Should I still include it in the vera implementation to make it consistent with the `lora` and `IA3` modules? I am also assuming the `com` and `com_inv` would also need to be re-included for `Vera` in order to allow for integration with the `LoRALayer` class

calpt:
> @calpt Regarding the paper: from my first readings it seems like it doesn't mention any scaling via a constant `alpha/r` or any gating. Should I still include it in the vera implementation to make it consistent with the `lora` and `IA3` modules? I am also assuming the `com` and `com_inv` would also need to be re-included for `Vera` in order to allow for integration with the `LoRALayer` class

yes, doesn't hurt to include these options even if not mentioned in the paper, unless this makes the implementation significantly more challenging

julian-fong:
I'll also include a new test module to test the `Vera` module similar to the IA3 and the Lora test modules soon

src/adapters/methods/lora.py
```diff
@@ -174,6 +179,111 @@ def forward(self, hidden_states: Optional[torch.Tensor], layer_input: torch.Tens
...
+            hidden_states = self.lora_dropout(hidden_states)
+
+        hidden_states = hidden_states @ self.vera_B @ lora_B @ self.vera_D @ lora_A
+
+        if self.use_gating:
+            gate = torch.sigmoid(self.gate(layer_input))
+            gate = torch.mean(gate, dim=1).unsqueeze(-1)
+            hidden_states = hidden_states * gate
+        else:
+            gate = None
```
calpt, line 250: as this is likely merged after #770, the same fix from there should be applied here

src/adapters/methods/lora.py
```diff
@@ -174,6 +179,111 @@ def forward(self, hidden_states: Optional[torch.Tensor], layer_input: torch.Tens
...
 
+class Vera(nn.Module):
+    def __init__(
+        self,
+        lora_A_shape,
+        lora_B_shape,
+        config: LoRAConfig,
+        gating_heads: int = 1,
+    ):
+        super().__init__()
```
calpt, line 190: we should also add an assert for composition mode "add" here (same as in LoRA init), just to make sure

src/adapters/configuration/adapter_config.py
```diff
@@ -499,6 +508,8 @@ class LoRAConfig(AdapterConfig):
     composition_mode: str = "add"
     init_weights: str = "lora"
     use_gating: bool = False
+    d: Union[bool, float] = None
+    b: Union[bool, float] = None
```
calpt, line 512: could we name these "vera_b" and "vera_d", to make more obvious what these are related to?

calpt: also why can these be bools? ie what happens when I set `d=True, b=True`?

src/adapters/methods/lora.py
```diff
@@ -90,6 +94,7 @@ def com_inv(self, weights: torch.Tensor, added: torch.Tensor) -> torch.Tensor:
         return weights - added * self.scaling
 
     def forward(self, hidden_states: Optional[torch.Tensor], layer_input: torch.Tensor):
+        print("triggered")
```
calpt, line 97: to remove?

src/adapters/methods/lora.py
```diff
@@ -174,6 +179,111 @@ def forward(self, hidden_states: Optional[torch.Tensor], layer_input: torch.Tens
...
+        lora_A = parameters["lora_A"]
+        lora_B = parameters["lora_B"]
+
+        if hidden_states is None:
+            hidden_states = layer_input
+
+        if getattr(self, "lora_dropout"):
+            hidden_states = self.lora_dropout(hidden_states)
+
+        hidden_states = hidden_states @ self.vera_B @ lora_B @ self.vera_D @ lora_A
```
calpt, line 243: shouldn't the order be reversed here? ie we matmul hidden states with lora_A -> vera_d -> lora_B -> vera_b, according to §3.1 (2) of the paper?

src/adapters/methods/lora.py
```diff
@@ -217,6 +331,10 @@ def add_adapter(self, adapter_name: str, layer_idx: int) -> bool:
                 lora_config,
                 gating_heads=self.get_n_heads(lora_config),
             )
+            # if we're using Vera, then set the adapter name into the Vera object
+            if lora_cls == Vera:
+                lora.set_vera_adapter_name(name=adapter_name)
```
calpt, line 336: feels a bit hacky to do this only for vera as the name is not specific to this type. what do you think of always passing the name directly to the `__init__` method of each module class (for all LoRA, Vera, IA3) and setting `self.name` directly there?
that might be cleaner long-term as we might want to use the name in LoRA as well in the future.

src/adapters/methods/lora.py
```diff
@@ -174,6 +179,111 @@ def forward(self, hidden_states: Optional[torch.Tensor], layer_input: torch.Tens
...
+        else:
+            gate = None
+
+        return hidden_states, gate
+
+    def set_vera_adapter_name(self, name):
+        self.name = name
+
+
+def init_shared_Vera_parameters(model_config, adapter_config, device):
```
calpt, line 258: nit: ideally lower-case "v" in the middle of method names

calpt (commented):
did a first review pass, thanks for working on this!
will review again once we have tests & docs added.

for adding tests, you might sync with @TimoImhof, since we have a larger test folder refactoring coming up here: https://github.com/adapter-hub/adapters/pull/740, so might make sense to directly base off that?

src/adapters/configuration/adapter_config.py
```diff
@@ -523,6 +534,27 @@ class IA3Config(LoRAConfig):
...
 
+@dataclass(eq=False)
+class VeraConfig(LoRAConfig):
+    """
+    Lora Config that applies vector-based random matrix adaptation. It adds
+    trainable matrices 'd' and 'b' while keeping the original LoRA matrices
+    frozen, random, and shared across layers. See more through their paper:
+    https://arxiv.org/pdf/2106.09685. Note that `r` will still be supplied
+    since we are still initializing decomposition matrices A and B.
+    The `composition_mode` parameter should also be set to `add`.
```
calpt, line 545: this doc still needs updating :)

src/adapters/methods/lora.py
julian-fong: I've thought about that idea as well but opted for now to implement this idea first since right now `lora` and `IA3` don't use `self.name`. I'll refactor it as you said. Thanks!

src/adapters/configuration/adapter_config.py
julian-fong: Good catch, i think this is a typo based on a previous idea I had which i scraped later.. thanks

src/adapters/configuration/adapter_config.py
```diff
@@ -535,6 +546,29 @@ class IA3Config(LoRAConfig):
     dtype: Optional[str] = None
 
 
+@dataclass(eq=False)
+class VeraConfig(LoRAConfig):
```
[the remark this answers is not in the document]
julian-fong: good catch!

julian-fong:
> This looks pretty good as far as I can tell! Did you get a chance to train a small vera adapter for testing?

Yeah I trained it and it works off of the first notebook on a simple example.

I've been struggling a bit on how to pass the vera tests (which i need to re-create since I had to reset to a previous commit to avoid the large amount of merge conflicts), will let you or @TimoImhof know if I ended up having some trouble

julian-fong:
@calpt I fixed most of the errors - I think the bottleneck one seems to be unrelated - and the last error is related to the adapters interface

lenglaender (approved):
Hey @julian-fong,

I fixed the remaining failing tests. There were 2 problems: the shape of LoRA A and B matrices used by Vera, and the hidden size check for encoder-decoder models and CLIP. Vera now uses the correct dimensions for the frozen LoRA A and B matrices, and I've added a hidden size check similar to the one we use for PHMLayers.

Once you've had a final look, we can merge this.

Thanks!

julian-fong:
Hi @lenglaender, Thanks for the help! I knew something was up with the dimensions and the hidden_size, but couldn't figure out why. It looks good, but only thing is that I disabled some of the tests inside various models eg: `tests/test_methods/test_on_encoder_decoder.py` for Vera. We should probably remove it and try the tests one more time?

julian-fong:
> Yes, enabling the tests for the Encoder-Decoder Models would be good. If it is enabled, only one test seems to fail: `tests/test_methods/test_on_encoder_decoder.py::Vera::test_load_full_model_Vera`. This has something to do with the shared parameters: We have a reference to the same parameters at multiple positions (in encoder.shared_parameters
> 
> And still missing is the documentation. If you could add the following this would be great:
> 
> * Description in docs/methods.md
> * Have it in docs/model_overview.md
> * And docs/overview.md
> * Add which composition blocks are supported in docs/adapter_composition.md
>   
>   * Are all compositions that LoRA supports also supported by VeRA? At least for averaging, this seems to be not the case as we must average the LoRA parameters in the shared parameters. This must happen in the model_mixin.py -> `average_adapter` method
> * Add it in the README "Implemented Methods" section
> 
> These are all the files I changed when I added prompt tuning (see #595 and #607 for what exactly has to be changed).
> 
> Ah and one last thing: You must need to add deleting the shared parameters in the `delete_adapter` method (in model_mixin.py) - just as we do for the PHM layer.

Thank you for the review! I will go ahead and do these changes and then ask for your review :)

julian-fong:
Hi @lenglaender 

I've addressed the majority of the changes you have asked for, and have just a few follow ups regarding some of the changes:

> Yes, enabling the tests for the Encoder-Decoder Models would be good. If it is enabled, only one test seems to fail: `tests/test_methods/test_on_encoder_decoder.py::Vera::test_load_full_model_Vera`. This has something to do with the shared parameters: We have a reference to the same parameters at multiple positions (in encoder.shared_parameters)

I am honestly not quite sure how to deal with this transformers error

```
RuntimeError: The weights trying to be saved contained shared tensors [{'decoder.bert.shared_parameters.dummy.lora_A', 'encoder.shared_parameters.dummy.lora_A', 'shared_parameters.dummy.lora_A'}, {'encoder.shared_parameters.dummy.lora_B', 'decoder.bert.shared_parameters.dummy.lora_B', 'shared_parameters.dummy.lora_B'}] that are mismatching the transformers base configuration. Try saving using `safe_serialization=False` or remove this tensor sharing.
```

>   * Are all compositions that LoRA supports also supported by VeRA?

Most of them I believe so, I'm not 100% sure about `MultiTask` or `Parameter Averaging`, so I've opted to leave them out for now incase I give them the check mark and they don't actually work haha

>  At least for averaging, this seems to be not the case as we must average the LoRA parameters in the shared parameters. This must happen in the model_mixin.py -> `average_adapter` method

In the `Vera` paper, the shared parameters are frozen and equal amongst all layers, would it be fine to just leave `self.base_model.shared_parameters[adapter_name]` as the average?

> Ah and one last thing: You must need to add deleting the shared parameters in the `delete_adapter` method (in model_mixin.py) - just as we do for the PHM layer.

Would the current method for PHM Layer work for Vera as well? I don't see an obvious reason why it couldn't be used for both

Let me know if you have any feedback on the above items or if you feel like I still am missing something! 

Thanks again ! :)

lenglaender:
Thanks for addressing most of the feedback, @julian-fong! I took care of the remaining points: Merging Vera adapters and the EncoderDecoder issues are now resolved, along with some minor cleanups (e.g. fixing the `average_adapter` method for `prompt_tuning`).

FYI: One significant change was in test_vera.py: I added `"shared_parameters.{name}."` to the `filter_keys` because Vera affects parameters there too, not just under `"loras.{name}."`. This was needed for the test to execute correctly for all parameters and helped surface a few other issues that are now fixed.

I will merge this PR now. Thanks again for working on this for the past months!

PR #767, opened 2024-12-10
Allow specifying adapter dtype in AdapterConfig

killershrimp:
Aims to fix https://github.com/adapter-hub/adapters/issues/766
Backwards compatible, since `dtype` defaults to `None` if not set in `AdapterConfig`.


calpt (approved):
Looks good to me, thanks for contributing!

src/adapters/configuration/adapter_config.py
```diff
@@ -499,6 +499,7 @@ class LoRAConfig(AdapterConfig):
     composition_mode: str = "add"
     init_weights: str = "lora"
     use_gating: bool = False
+    dtype: Optional[str] = None
```
calpt, line 502: one minor request: could you add the dtype attribute to the python doc of the changed classes? thanks!

Issue #760, opened 2024-11-19
Different outputs when upgrading from adapter-transformers with LoRA


jblamare:
## Environment info

- `adapters` version: 1.0.1
- `transformers` version: 4.45.2
- Platform: Linux-5.15.0-91-generic-x86_64-with-glibc2.35
- Python version: 3.10.12
- Huggingface_hub version: 0.26.2
- Safetensors version: 0.4.5
- Accelerate version: not installed
- Accelerate config: not found
- PyTorch version (GPU?): 2.5.1+cu124 (True)
- Tensorflow version (GPU?): not installed (NA)
- Flax version (CPU?/GPU?/TPU?): not installed (NA)
- Jax version: not installed
- JaxLib version: not installed
- Using distributed or parallel set-up in script?: No
- Using GPU in script?: Yes
- GPU type: NVIDIA RTX A6000

## Information

Model I am using (Bert, XLNet ...): `google/flan-t5-small`

Language I am using the model on (English, Chinese ...): English

Adapter setup I am using (if any): LoRA

The problem arises when using:
* [ ] the official example scripts: (give details below)
* [x] my own modified scripts: (give details below)

The tasks I am working on is:
* [ ] an official GLUE/SQUaD task: (give the name)
* [x] my own task or dataset: (give details below)

## To reproduce

I have two environments:
- env1
```
adapter-transformers==3.1.0
torch==1.13.1
```
- env2
```
adapters==1.0.1
torch==2.5.1
transformers==4.45.2
```
I have some input_ids 
```
input_ids = [262, 4, 4815, 10, 8668, 3, 5359, 27415, 5332, 3430, 276, 3577, 20186, 11951, 8472, 11359, 4209, 205, 20931, 23936, 3388, 27447, 8015]
```
I have a model checkpoint `checkpoint.pth` which has the T5 weights plus a LoRA adapter and was saved in env1:
```
with open("checkpoint.pth"), "wb") as f:
    torch.save(model.state_dict(), f)
```

From there I want to make sure I can load the model, run inference, and get the same outputs in env2. But the outputs are different. I run the following experiments:
1. Create a T5 model, add the empty LoRA adapter, run inference - env1 and env2 get to the same output
2. Create a T5 model, load the non-LoRA weights, run inference - env1 and env2 get to the same output
3. Create a T5 model, add the LoRA adapter, load all the weights, run inference - env1 has the right output but env2 is different.

[...]

PR #770, opened 2024-12-22
Fix default Lora/ (IA)^3 scaling in forward

calpt:
Resolves issue described in #760.

**IMPORTANT**: this fix restores weights compatibility with adapter-transformers. Compatibility to previous adapters versions is kept via a compat patch.

## Details

The current implementation of LoRA/ (IA)^3 in `adapters ` versions < 1.1.0 does not correctly implement adapter states scaling via the LoRA `alpha` attribute, effectively ignoring `alpha` and always applying a scaling of 1.0.
This PR restores the correct original behavior as found in adapter-transformers/ original LoRA implementation.

As this change breaks all adapters pre-trained using `adapters` versions 0.1.0 - 1.0.1, a backward compatibility patch is introduced that automatically sets `alpha = r` for LoRAs for adapters that were trained using affected versions. This ensures all previous adapters continue to behave exactly as trained (ie give the exact same output using newer versions).

### DIFF
--- a/src/adapters/loading.py
+++ b/src/adapters/loading.py
@@ -371,7 +371,7 @@ class AdapterLoader(WeightsLoader):
 
     def _fix_backward_compat(self, config):
         # Fix error in previous versions for LoRA/ (IA)^3
-        if config["version"].startswith("adapters.") and Version(config["version"][9:]) < Version("1.1.0"):
+        if config.get("version", "").startswith("adapters.") and Version(config["version"][9:]) < Version("1.1.0"):
             if (
                 config["config"].get("architecture", None) == "lora"
                 and config["config"]["r"] != config["config"]["alpha"]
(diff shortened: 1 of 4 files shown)


src/adapters/loading.py
```diff
@@ -368,6 +369,19 @@ def _rename_legacy_weights(self, k):
             k = k.replace(old, new)
         return k
 
+    def _fix_backward_compat(self, config):
+        # Fix error in previous versions for LoRA/ (IA)^3
+        if config.get("version", "").startswith("adapters.") and Version(config["version"][9:]) < Version("1.1.0"):
```
TimoImhof, line 374:
```suggestion
        ADAPTER_PREFIX = "adapters."
        MIN_VERSION = Version("1.1.0")

        version = config.get("version", "")
        if version.startswith(ADAPTER_PREFIX) and Version(version[len(ADAPTER_PREFIX):]) < MIN_VERSION:
```

TimoImhof: Would remove unnecessary double lookup in config and has no hardcoded version suffix

TimoImhof (approved):
Thank you for working on this!
This looks good to me, I only provided some minor suggestions that could be included.

PR #782, opened 2025-01-13
Upgrade Transformers to 4.48.x

TimoImhof:


src/adapters/models/beit/mixin_beit.py
```diff
@@ -9,7 +9,7 @@
 from ...utils import patch_forward
 
 
-class BeitSelfAttentionAdaptersMixin:
+class BeitAttentionAdaptersMixin:
```
calpt, line 12: why is this renamed?

src/adapters/models/llama/modeling_llama.py
```diff
@@ -52,166 +48,31 @@ class LlamaAttentionWithAdapters(LlamaAttentionMixin, LlamaAttention):
...
-        # batch_size x seq_length x head_dim x hidden_dim
-        # therefore we just need to keep the original shape
-
         # Loosen constraint on batch_size to allow parallel adapter composition
-        query_states = query_states.view(-1, q_len, self.num_heads, self.head_dim).transpose(1, 2)
-        key_states = key_states.view(-1, q_len, self.num_key_value_heads, self.head_dim).transpose(1, 2)
-        value_states = value_states.view(-1, q_len, self.num_key_value_heads, self.head_dim).transpose(1, 2)
+        query_states = query_states.view(-1, q_len, self.config.num_attention_heads, self.head_dim).transpose(1, 2)
+        key_states = key_states.view(-1, q_len, self.config.num_key_value_heads, self.head_dim).transpose(1, 2)
+        value_states = value_states.view(-1, q_len, self.config.num_key_value_heads, self.head_dim).transpose(1, 2)
```
lenglaender, line 67: Why do we have different lines here for computing query_states etc. compared to the computation in the forward functions of HF's forward function: 
https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py#L267-L268

Can't we copy this part of HF's code too and then only add our `match_attn_matrices_for_parallel` and `adjust_tensors_for_parallel`? Currently, a huge chunk of code is different to hugging face (and thus marked as AH changes)

src/adapters/models/beit/mixin_beit.py
TimoImhof: Renamed it without a deeper reason behind it, I reverted it to the original.

src/adapters/models/llama/modeling_llama.py
TimoImhof: In principle, I also prefer to keep the modifications minimal, but I don't know how we should implement this otherwise.
The HF implementation assumes that only  `num_attention_heads` and `num_key_value_heads` are different and that the shapes of the keys, queries, and values are identical otherwise.

However, when using an adapter composition, this assumption does not hold; we modify the shapes of queries and values and have an additional "unknown" shape variable.

So, without the relaxation, I don't think we can determine the correct batch size.

lenglaender: Yeah okay, then let's proceed with merging the PR as it is :+1:

PR #790, opened 2025-02-04
[ENH] Implements DoRA

julian-fong:
Paper: https://arxiv.org/pdf/2402.09353
Github: https://github.com/NVlabs/DoRA/

draft pr for `DoRA`, still need to write the compose function


julian-fong:
@calpt I've written out the compose function, using https://github.com/NVlabs/DoRA/blob/7ee989695252cb0f4f7579182c81581aa75139a7/commonsense_reasoning/peft/src/peft/tuners/dora.py#L411 as an reference point.

Using the first notebook as a testing standpoint, it does train, however the performance is quite poor (does not outperform LoRA by a decent amount). maybe there is something wrong with my implementation (i'm also curious as to why they do norm_scale-1) 

It also does not seem straightforward/simple to write out the `delta_w` methods and the `inv_com` method, as i believe it would require the pretrained weights W_o as an input for the calculation.

An initial review is appreciated :)

src/adapters/methods/lora.py
```diff
@@ -432,6 +442,60 @@ def init_shared_vera_parameters(lora_A_shape, lora_B_shape, adapter_config, devi
     return parameters
 
 
+def init_m_dora_weight(lora_b_shape, dtype):
+    m = nn.Linear(1, lora_b_shape[0], bias=False, dtype=dtype)
+
+    return m
```
calpt, line 448: according to the paper, we'd also want to initialize m = ||W_0||, no?

src/adapters/methods/lora.py
```diff
@@ -432,6 +442,60 @@ def init_shared_vera_parameters(lora_A_shape, lora_B_shape, adapter_config, devi
...
+    return result
+
+
+def compute_dora_add_com_inv(
+    weights: torch.Tensor, added: torch.Tensor, m: torch.Tensor, norm: torch.Tensor
+) -> torch.Tensor:
+    """This function returns the required weights necessary
+    to compute the inverse composition where `composition_mode` == add.
+    """
+    result = weights - weights * norm.unsqueeze(1) / m.weight - added
```
calpt, line 495: similar to above, I think we should try to make this more readable

docs/methods.md
```diff
@@ -295,6 +295,26 @@ Using the `VeraConfig`, you can specify the initialization of the scaling vector
...
+
+$$ h = W_{0}x + $m$\frac_{W_{0}x + BAx}{\norm{W_{0} + BA}} $$
+
+.. note::
+    Unlike the previous LoRA tuning methods, DoRA is a method *does not* have its own config but instead is accessed via the parameter `use_dora` in other LoRA-like configs. See below for a quick example.
+```
+
+_Example_:
+```python
+config = LoRAConfig(use_dora = True)
```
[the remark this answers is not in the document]
calpt: Also might make sense to add "dora" and "dvora" to `ADAPTER_CONFIG_MAP` in `adapter_config.py`

src/adapters/methods/lora.py
julian-fong: i saw it in the paper as well but when going through thier code: https://github.com/NVlabs/DoRA/blob/7ee989695252cb0f4f7579182c81581aa75139a7/commonsense_reasoning/peft/src/peft/tuners/dora.py#L357 according to this line, they initialize `m` like this unless i'm reading it incorrectly

src/adapters/methods/lora.py
```diff
@@ -432,6 +442,60 @@ def init_shared_vera_parameters(lora_A_shape, lora_B_shape, adapter_config, devi
...
+    """This function calculates the dora update.
+
+    In the paper, the dora update delta_w is calculated as follows:
+    m * (w_0x + BAx) / ||w_0 + BA||c
+    """
+    norm_scale = m.weight.view(-1) / norm
+    scaled_weights = (norm_scale - 1) * weights
+    scaled_lora = norm_scale * added
+    result = scaled_weights + scaled_lora
+    return result
```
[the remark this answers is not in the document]
julian-fong: Thats right, the idea here for the method `compute_dora_deltaw` was to formulate the 'BAx' that goes into the `Vera` or `LoRA` layers. I tried to formulate it this way so that we could still utilize the original `com` method and keep as much of the original code intact. the method also isn't exactly too extensible since its designed specifically for the "W_oX + BAx" com method

src/adapters/methods/lora.py
calpt: they seem to initialize the weights with the norm at a different place: https://github.com/NVlabs/DoRA/blob/7ee989695252cb0f4f7579182c81581aa75139a7/commonsense_reasoning/peft/src/peft/tuners/dora.py#L242-L244

src/adapters/methods/lora.py
```diff
@@ -432,6 +451,59 @@ def init_shared_vera_parameters(lora_A_shape, lora_B_shape, adapter_config, devi
...
+    m * (w_0x + BAx) / ||w_0 + BA||c
+    where: 
+        - m is the magnitude parameter
+        - norm is the column-wise norm i.e ||w_0 + BA||c
+        - orig_result is the layer output w_0x
+        - added is the LoRA update BAx
+    """
+    norm_scale = m.weight.view(-1) / norm
+    results = orig_result + (norm_scale - 1) * orig_result
+    results += (norm_scale * added) * scaling
```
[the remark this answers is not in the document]
julian-fong: right, this is equivalent and honestly a better equation to look at from a dev perspective, will make the changes

src/adapters/methods/lora.py
```diff
@@ -432,6 +451,58 @@ def init_shared_vera_parameters(lora_A_shape, lora_B_shape, adapter_config, devi
...
+    and the LoRA matrices
+
+    eg: for LoRA, it would be ||w_0 + BA||c
+    """
+    return torch.linalg.norm(weights + added, dim=1)
+
+
+def compute_dora_deltaw(
+    orig_result: torch.Tensor, added: torch.Tensor, m: torch.Tensor, norm: torch.Tensor, scaling: float = 1.0
+) -> torch.Tensor:
```
calpt, line 466:
```suggestion
def compute_dora_deltaw(
    orig_result: torch.Tensor, added: torch.Tensor, m: torch.Tensor, norm: torch.Tensor
) -> torch.Tensor:
```

scaling not needed anymore

PR #796, opened 2025-02-17
Unlock version of transformers

devin-astrumu:
Is there any reason why we lock transformers version to 4.47.1 ? This conflicts with many packages, like latest vllm.


XBeg9:
duplicate of #782

PR #797, opened 2025-03-07
Adapter Interface: Validate Values

lenglaender:
This PR introduces a validation for the values specified in the adapter interfaces. We now check that the layers specified actually exist and if not, we provide a detailed error to the user so that they know what went wrong.

Before this PR, when users specify layers in the adapter interfaces that don't exist in the model, there was no error thrown. This left users unaware that their specified values were invalid.

### DIFF
--- a/src/adapters/wrappers/model.py
+++ b/src/adapters/wrappers/model.py
@@ -30,6 +30,9 @@ SPECIAL_MODEL_TYPE_TO_MODULE_NAME = {
 }
 
 
+_INTERFACE_ERROR_TEMPLATE = "AdapterInterface: '{layer_name}' is set to '{layer_value}' but this value is not found in the {parent_name}. See https://docs.adapterhub.ml/plugin_interface.html for more information."
+
+
 def get_module_name(model_type: str) -> str:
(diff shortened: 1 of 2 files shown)


src/adapters/wrappers/model.py
```diff
@@ -167,3 +169,77 @@ def new_init(self, config, *args, **kwargs):
...
+    Args:
+        base_model: The base model to validate against
+        interface: The adapter interface to validate
+
+    Raises:
+        ValueError: If any specified path is not found in the model
+    """
+
+    if not multihasattr(base_model, interface.model_embeddings):
+        raise ValueError(
```
calpt, line 187: nit: I think we could use a common template for this error message as it is roughly the same in each instance

calpt (approved):
Thanks for adding!

src/adapters/wrappers/model.py
lenglaender: I thought so too, but then again thought that adding that would introduce slightly more complexity to the code so originally didn't do it. But I've implemented the template now. Thanks for the review!

PR #802, opened 2025-04-21
Add plugin interface notebook & Qwen 2/ 2.5 interface

calpt:
- Adds a demo notebook showcasing how to work with the plugin interface
- Adds a default interface for Qwen2/ Qwen 2.5
- Automatically saves custom interfaces along with the adapter

PR #804, opened 2025-04-24
HotFix: Remove accidental pdb.set_trace() call from test file

TimoImhof:
This PR removes a hidden pdb.set_trace() call in test_adapter_composition.py:

``` python
__import__("pdb").set_trace()
```

### Problem
This line caused the test runner (pytest) to pause execution and wait for user input during test discovery or execution. Since it's not in an interactive terminal (especially when running via VS Code or CI), it appears as if the tests are hanging or failing silently.
It looks like this was a leftover debugging breakpoint that was accidentally committed.

### Fix
The line has been removed to allow the test suite to run normally without interactive interruption.
After removing this, test discovery and execution complete successfully.


TimoImhof:
Was then included in #802

PR #805, opened 2025-05-02
Upgrade transformers to 4.51.3

TimoImhof:
This PR upgrades transformers to v4.51.x.

Required changes in the adapters code base:
- Properly skip sdpa inference tests for CLIP
- Remove deprecated tpu availability method from example scripts
- Update weight initialization for BertFusion
- Avoid tensor allocation on meta device for LoRA and REFT parametrization

### DIFF
--- a/examples/pytorch/dependency-parsing/utils_udp.py
+++ b/examples/pytorch/dependency-parsing/utils_udp.py
@@ -25,14 +25,9 @@ from transformers import (
     Trainer,
     TrainerCallback,
     TrainingArguments,
-    is_torch_tpu_available,
 )
 from transformers.trainer_utils import PredictionOutput
 
-
-if is_torch_tpu_available():
-    import torch_xla.core.xla_model as xm
-    import torch_xla.debug.metrics as met
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


lenglaender:
I looked more into the REFT problem:

It seems the way Hugging Face Transformers loads models has changed quite a bit, especially in their `_load_pretrained_model` method in `modeling_utils.py`. With these new changes, our `reft.py` (when `orthogonality=True`) is running into issues.

What I discovered is: When we use `nn.utils.parametrizations.orthogonal`, it is called with `use_trivialization=True` _(which we want for better convergence, like the original REFT paper)_. When called with `use_trivialization=True`, it creates an extra `base` buffer (that is the `...projection.parametrizations.weight.0.base`) inside the projection layer.

The problem is, with Transformers' new way of initializing models, this `base` buffer isn't properly set up or recognized during the initial meta-model(?) creation. They now call `missing_keys, unexpected_keys = _find_missing_and_unexpected_keys(` early and the `state_dict()` doesn't yet contain the `base` buffer at this point. But, our saved checkpoint does have this `base` key. In previous versions of Transformers, this loading worked completely different and there was no `_find_missing_and_unexpected_keys` function.

One way around this is to set `use_trivialization=False` in the `orthogonal` call (and then we need to also remove our `pre_save_adapters` method since there's no `base` buffer). If I do that, all our tests pass: it works because there is no buffer. But the original REFT paper used `use_trivialization=True` because it helps with convergence, so we should stick with that too.

I'm a bit stumped on how to fix this while keeping `use_trivialization=True`. I know you, @calpt , and @TimoImhof  have spent hours on this too, but maybe these insights can help you. The only thing I can think of is trying to somehow patch or overwrite parts of `_load_pretrained_model`, but I'm not quite sure on what we would even need to change there. Honestly, I'm not entirely sure why this worked fine with older versions of Transformers.

calpt (approved):

lenglaender (approved):

PR #812, opened 2025-05-25
Fix invalid .json for adapter notebooks

TimoImhof:
While scanning through the notebooks, @julian-fong encountered the following error:
image

Given the instructions in the error message on how to fix the issue I took option two and removed the metadata.widgets key with its content from the affected notebooks. As a result the notebooks can be rendered again normally.


TimoImhof:
> Looks good to me! I checked it with the third solution proposed here and came to the same changes: https://github.com/orgs/community/discussions/155944#discussioncomment-12845735
> 
> @julian-fong Timo didn't clear all the cell outputs but only the widgets that are wrongfully parsed by GitHub (since ~ beginning of April). As far as I can tell, removing the cells that we removed doesn't seem to affect the visual appearance at all (since its metadata).

I added the script I used for fixing the problem as a new comment: https://github.com/orgs/community/discussions/155944#discussioncomment-13465574

